Issue #10, opened 2013-03-15
could support multiple language or custom language


bastengao:
Hi, i am a Chinese. I try `intro.js`, it is cool and useful and i want to use it on our production, but it is regretful that it not supports custom language now, because our customers are also Chinese.

creativeprogramming:
+1, but you can help us in this also in a simple way: please document how to relabel next/prev buttons, is it possible with setOption? 

Anyway awesome project! congrats.

PR #60, opened 2013-03-21
Added ability to set next, skip, and previous button text via new .setoptions({}) method

sayhiben:
This is intended to fix issue #10

This addition allows for the labels of the buttons to be set by calling:

``` javascript
introJs().setoptions({nextLabel: 'This Way', skipLabel: 'Go Away', prevLabel: 'That Way'}).start();
```

Or, individually like

``` javascript
introJs().setoption('nextLabel', 'This Way').start();
```

This adds two methods on the public object:

``` javascript
.setoptions({key: 'val', key2: 'val2'})
.setoption('key', 'val')
```

And one method to the private object:
`_mergeOptions(obj1, obj2)`

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -18,6 +18,12 @@
    */
   function IntroJs(obj) {
     this._targetElement = obj;
+    
+    this._options = {
+      nextLabel: 'Next &rarr;',
+      prevLabel: '&larr; Back',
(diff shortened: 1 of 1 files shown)


creativeprogramming:
please also the 'Done' button label

PR #80, opened 2013-03-28
start intro for group of elements

Delysid:
`data-intro-group` option of element specifies intro group, which you can launch by `introJs.start(group_name)`


piotr-cz:
Seems that this feature is not present in the v2 anymore

afshinm:
@piotr-cz will have a look

piotr-cz:
Sorry, this feature has never been merged (see [v1.1.1](https://github.com/usablica/intro.js/blob/v1.1.1/intro.js)).

I'm planning to work around it:

1. I store intro text in in other, group data attribute (`data-intro-namespace="menuitem", data-intro-menuitem="Click here for fireworks"`)
2. _before_ introJs Start copy it's contents to data-intro attribute
3. _after_ introJs end remove data-intro attributes

tnchuntic:
is this implemented already? I couldn't make it work on my end.

afshinm:
@bozdoz what do you think? I believe it should be easy to rebase / reimplement.

bozdoz:
Agreed. I can take a look at this shortly.

xmflswood:
+1

dspoonia7:
+1

PR #131, opened 2013-05-17
Separate content + HTML

alvarotrigo:
I have made use of jQuery as I am using it in my application, but I'm sure you can do it with pure JavaScript.

I have managed to insert HTML in the tour popups as well as completely separate the tour content from the page layout.
`data-intro` attributes are no longer needed. Instead, a list containing information about the elements to take part in the tour and its text is needed as well as the floating box position.

`data-step` is not either needed. It will follow the list elements order.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -67,22 +67,35 @@
       }
 
     } else {
-      //use steps from data-* annotations
+      //use steps from the list #tourContent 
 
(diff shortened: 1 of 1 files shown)


danielgolden:
Please revive this issue. I would absolutely love to be able to specify my content in html somewhere else on the page. Tough to do right now with either having to place it in an attribute or in json (where I don't have access to php variables).

PR #208, opened 2013-10-03
Skip hidden elements

Syerram:
Currently if the element is hidden, intro.js will still show the help but at top/left corner. Usually we hide/show elements depending on what part of screen they are in and also depending on their security creds. 

So this is a serious issue since the help text cannot be shown to people without creds.


afshinm:
manually merged. thanks all.

PR #236, opened 2014-01-27
fix issue with single-step pages (closes #229)

### LINKED ISSUE #229: The problem with multi-page
I have single one step on page 1, but there is only one button - skip, and by clicking on it the transition occurs to page two.

adrianp:
- Fixes issue with "Done" label not shown on pages with a single
  data-step.

Signed-off-by: Adrian-Tudor Panescu adrian.tudor.panescu@cern.ch

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -582,11 +582,11 @@
       _placeTooltip.call(self, targetElement.element, tooltipLayer, arrowLayer);
     }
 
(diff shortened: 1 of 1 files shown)


curiousPup:
this is still an issue

fritx:
still throw such error when I had only one step in array (but I'm using multiple instance via intro.js-react)

PR #240, opened 2014-02-14
fixes #96

### LINKED ISSUE #96: navbar fixed top, positioning issue
Hello,

First of all, thank you very much for this amazing plugin!
I'm using Twitter bootstrap for the layout of my site and I put one of my steps in a navbar fixed top. Making the the number step cut off... (see princtscreen).

Screen Shot 2013-04-13 at 19 22 12
Any chance we could also change the position of the step number for such cases?

Thanks,
Denis

sharpnk:


ladigitale:
Listen to onchange event :
- clone the target element (targetElement.clone(true)),
- measure target element by using getClientBoundingRect
- set clone dims (top, left, width, height)
- set z-index to 10000000
- mask target element
- add clone to body

don't forget to remove clone et show target element on next change and on exit

vhmth:
@ladigitale that's the only thing that did it for me. I clone, get the position from the introjs tooltip layer, and then append the clone to the body and apply the position stylings.

PR #254, opened 2014-03-14
Add data-scroll-to-element attribute for changing the scrollToElement option

Urigo:


buccolo:
Nice, I was going to do something similar :+1:

PR #259, opened 2014-03-28
fixes #194 by changing the CSS class for "Done" so it's not the same as ...

### LINKED ISSUE #194: Is it possible to disable 'Skip' button?
In certain situations, I want to make sure that the users complete the tour, so an option for disabling the 'Skip' button would be nice.

Can we do this in intro.js at the moment?

waynn:
..."Skip"

I'm not sure if this is the best way to fix it, but this is how I've done it in my code, and it gives users the choice of changing CSS for Skip vs. Done.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -687,6 +687,7 @@
       skipTooltipButton.innerHTML = this._options.skipLabel;
     } else if (this._introItems.length - 1 == this._currentStep || this._introItems.length == 1) {
       skipTooltipButton.innerHTML = this._options.doneLabel;
+      skipTooltipButton.className = 'introjs-button introjs-donebutton';
       prevTooltipButton.className = 'introjs-button introjs-prevbutton';
       nextTooltipButton.className = 'introjs-button introjs-nextbutton introjs-disabled';
     } else {


rinogo:
@afshinm - Thanks for merging this!  I just saw the [concerns](https://github.com/usablica/intro.js/pull/259#issuecomment-39550393) expressed by @dzrodrigues... Do those need to be addressed, or are they irrelevant?  I'm happy to submit another PR if they need to be addressed.

rinogo:
@dzrodrigues, would a simple fix to your concerns simply be to modify the CSS instead?  e.g. to apply the same styles to the new `.introjs-donebutton` that are applied to `.introjs-skipbutton`?

afshinm:
@rinogo I will test them myself.

rinogo:
By the way, I just want to say "thanks" again for the PR, @waynn, and for the merge, @afshinm.  This allows us to do something like the following (just a simple two-step test):

image

image

We need to keep the UI simple and congruent with the existing UI since our users are not particularly computer savvy and need everything to be as simple as possible.  Anyway, thanks again, and have a happy new year!

afshinm:
I like that idea @rinogo. I'd like to have it as an additional option in the library. Could you send a PR?

rinogo:
@afshinm Which idea - https://github.com/usablica/intro.js/pull/259#issuecomment-269792066 or https://github.com/usablica/intro.js/pull/259#issuecomment-269794576 ?

https://github.com/usablica/intro.js/pull/259#issuecomment-269794576 was accomplished by following the same approach used in [this Codepen](https://codepen.io/jesperkc/pen/VeaLaB).

PR #270, opened 2014-04-11
Better management of up/down scroll

danmaz74:
With this change the position and (estimated) size of the tooltip is considered before deciding if a scroll is needed, and how much scrolling to do. This fixes the bug where many times the tooltip is not visible after a scroll.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -723,21 +723,41 @@
       parentElm = parentElm.parentNode;
     }
 
-    if (!_elementInViewport(targetElement.element) && this._options.scrollToElement === true) {
-      var rect = targetElement.element.getBoundingClientRect(),
-        winHeight=_getWinSize().height,
-        top = rect.bottom - (rect.bottom - rect.top),
-        bottom = rect.bottom - winHeight;
-
-      //Scroll up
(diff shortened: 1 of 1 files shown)


analog-nico:
This may fix issue #503. Check out the jsFiddle there.

afshinm:
@analog-nico I have added a scroll padding option which is 30 by default. Can this solve your issue?

analog-nico:
@afshinm Unfortunately not. The issue is that the code still doesn't consider the size of the tooltip.

For example these two steps refer to elements that are both in the viewport. Thus no scrolling takes place and the second tooltip is cut off:

screen shot 2016-09-27 at 11 50 11

On big screens this is often not an issue. But since we nowadays need to build everything response there is a lot of tweaking required. In the above case it would mean that I need to attach the second tooltip at the bottom. I had to apply this tweak for many steps in my onboarding tour. For a much better setup experience it would be great if intro.js would do this automatically like in this pull request.

adfink:
I'm having the same issue described above.  Is there a reason this pr was not merged in?

russellmann:
Also looking for this functionality.

PR #286, opened 2014-06-02
Do not apply introjs-relativePosition on elements with position fixed 

vincentlaucy:
as title as when `position:fixed` became `position:relative` layout breaks


pavlokitdev:
+1

afshinm:
manually merged. will release the v2.0.1 soon.

afshinm:
thanks all

Issue #281, opened 2014-05-14
Error when clicking on Bullet when only a single item is showing


timherby:
The bullets show up when there is only one intro item, which is probably a bug unto itself.  
![screenshot 2014-05-13 23 08 54](https://cloud.githubusercontent.com/assets/4109428/2967444/518a4b94-db2e-11e3-81df-e9163c4aa4a6.png)

If you click on the bullet, since there is no previous item, the code on line https://github.com/usablica/intro.js/blob/master/intro.js#L690 throws an exception:

```
Uncaught TypeError: Cannot set property 'className' of null application.js:49302
_showElement application.js:49302
_nextStep application.js:48898
_goToStep application.js:48864
introJs.fn.IntroJs.goToStep application.js:49563
_showElement.anchorLink.onclick
```

I would think the easiest way to fix would be to disable bullets if there is only one item in the list.

rvenk:
I used the following css to make the bullet invisible when there is only 1 introjs item:
```css
.introjs-bullets li:only-child {
  display: none;
}
```

PR #293, opened 2014-06-19
Check that previous and next buttons exist before trying to add classes to them.

nickpyren:
In addition, some of the logic of setting the classes for the next, previous, and innerHTML for the skip buttons were reworked to make it more readable and simple. This is a solution to the issue in  #281.


bgruening:
@afshinm you should merge this one by hand.

PR #299, opened 2014-07-15
Add new function `goToDataStep`

doits:
This allows to go to a specific step with the specific `data-step`
attribute. This differs from `goToStep` in the way that `data-step`
does not have to be continuous to pick the desired element.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,32 @@ introJs().goToStep(2).start(); //starts introduction from step 2
 
 -----
 
+###introJs.goToStepNumber(step)
+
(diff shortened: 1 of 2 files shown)


ir-fuel:
Will this ever be available?

afshinm:
sorry for this delay, could you please rebase with master? thanks a lot.

doits:
The branch for this PR is long gone. I created a new PR #665 on master. Did not test if it works (just applied the patch) and the readme entry gone missing though.

PR #301, opened 2014-07-17
Circumvent fixed position stacking issues by using transparent overlay

jonkafton:
If you're interested (though changes visual behaviour).

- Make overlay transparent (use only to capture click events to exit)
- Make helper transparent (used to highlight target with box shadow)
- Remove any logic to reposition target or its parents. Assumes interactivity with target is undesirable whilst open.
- Position fixed helper if target or any of its parent are fixed.
- Remove button styles on assumption that apps will generally want to use global styles for buttons.

Worked really well for us with both static and fixed targets...

image

image


0xPr0xy:
Or this might be an option: [Guide.js](https://github.com/Dozyatom/guide.js) has a different implementation, that does not suffer from the z-index with fixed elements issue in WebKit

PR #335, opened 2014-09-25
Number positioning

bostikforever:
Sometimes it's desirable to have control over the positioning of the step numbers. (Again, this would be desirable in the case of elements in the corners). This set of changes adds that feature. Here's a demo: http://jsfiddle.net/87BLy/7/


raduluchian:
This is a very useful feature. A very common scenario is when someone is using Intro.js to highlight something in the navbar. In such a case the number is not visible if placed at the top.

bozdoz:
Looks really good to me, and I think it would solve a few issues, @afshinm

PR #346, opened 2014-10-02
Updated README for consistency

richguan:
Removed 'to' and added period to match the rest of the document


bozdoz:
Don't think this is still an issue.

PR #360, opened 2014-10-22
using scroll to animation

leejordan:
This causes the viewport to animate scroll to the next step where appropriate. I wrote it for my own implementation of intro.js. Hope you find it useful.

It uses the easing functions provided here: http://robertpenner.com/easing/. It would be nice to extend intro.js to allow the user to pass an option in to choose their easing type maybe? For me however quad easing is perfect.

Feel free to let me know if you find any bugs.


michalmikolajczyk:
Thanks!
I needed that functionality, so used what @leejordan provided.
My only concern is performance, it seems to be a bit edgy?

PR #367, opened 2014-11-06
Call oncomplete callback when Enter is hit with Done focused (fixes #366)

### LINKED ISSUE #366: oncomplete callback isn't called when hitting Enter with Done focused
Example here: http://jsfiddle.net/LpLtnyrs/
On step 2 the Done button is initially focused. Hitting Enter completes the intro but the oncomplete callback isn't called.
If there's just one step, the Done button isn't focused (that's another (minor) issue), so at least 2 steps are required to demonstrate the problem.

yuvale:
Fixes issue #366.
Also fixes the issue that Done isn't focused if there's just one step.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -214,7 +214,7 @@
             _previousStep.call(self);
           } else if (target && target.className.indexOf('introjs-skipbutton') > 0) {
             //user hit enter while focusing on skip button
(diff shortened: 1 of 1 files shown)


ericenns:
Please close this was fixed in PR #412

PR #375, opened 2014-11-27
Feature/element late binding

rnicholus:
This simple adjustment allows a function to be specified for an `element` property on a `step` object.  This allows for the element to be shown in the next/pending step to be evaluated as late as possible to account for dynamically rendered/added elements that are part of a tour.  

The default behavior of introjs seems to involve resolving all element selectors into `HTMLElement`s before the tour begins.  This doesn't work well with SPA and pages where elements are added after the DOM has loaded.

We are currently making use of this in an internal project, and this change appears to be necessary to properly include angular directives in the tour that contain contain elements that are conditionally rendered based on some async task result (ajax request, for example).

If there is anything that I can adjust here to make this more suitable (if it isn't suitable already), let me know.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -306,6 +306,9 @@
     }
 
     var nextStep = this._introItems[this._currentStep];
+    if (typeof nextStep.element === 'function') {
+      nextStep.element = nextStep.element();
+    }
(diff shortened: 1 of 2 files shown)


rnicholus:
preeeety sure this isn't going to be merged. closing

PR #376, opened 2014-11-30
added callback support for positioning...

janposselt:
...when using setOption or when defining steps programmatically. See https://github.com/janposselt/intro.js/blob/introduce-position-callback/example/custom-position/index.html for working example.


bozdoz:
Cool idea.

PR #377, opened 2014-12-03
Fix documentation error

zanderle:
There was an error in the documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ Set a group of options to the introJs object.
 
 **Example:**
 ```javascript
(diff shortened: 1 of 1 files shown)


zanderle:
I see. It might be useful to put that in the documentation then?

bozdoz:
Does not seem to apply to the current state of the repo.

zanderle:
Ok, I can close then. After all the PR is 3 years old...

bozdoz:
Thanks, @zanderle, and I do realize how untimely this is.  I just became a regular contributor a month or so ago, and am trying to clean things up.  :smiley:

PR #379, opened 2014-12-03
fixes #378 - bug( tooltip left offset calculation ) 

### LINKED ISSUE #378: Left-offset is miscalculated when the target element is thinner than the tooltip
When the width of the target element > the width of the tooltip the left-offset is miscalculated.  The offset should be the absolute value of the difference in half their widths.

Live demo:  http://jsfiddle.net/v4u6mo0y/

screenshot 2014-12-03 13 45 06

screenshot 2014-12-03 13 45 12

nmall:
fixes #378 
fixing when position is bottom and target element is thinner than tooltip.

screenshot 2014-12-03 13 50 39

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -367,7 +367,7 @@
     var referenceLayer = targetElement.querySelector('.introjs-tooltipReferenceLayer');
     if (referenceLayer) {
       referenceLayer.parentNode.removeChild(referenceLayer);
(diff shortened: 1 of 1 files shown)


thotmx:
@nmall I'm using your fix and it works really well. Thanks.

PR #384, opened 2014-12-19
Batman API for introJs - Trigger helpers and use them as steps-activators

nickbalestra:
You can see it in action here: http://nickbalestra.github.io/intro.js/example/batman/

About:
Inspired by a great post by @einkoenig: [batman onboarding](https://medium.com/@einkoenig/batman-onboarding-999d19f0cab9). Batman for introJs automate/facilitate via a simple API the job of adding contextual helpers over the UI so that users:
1. Don't need to go trough all the introduction when they just need to learn about a specific step.
2. Can get back to the wizard when they really needed it.
3. Feel more in control as they know where/what will trigger a step in the introduction.

As this is an ongoing experiment, I would love to hear what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+.DS_Store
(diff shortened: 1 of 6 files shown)


afshinm:
This is super cool but unfortunately I can't merge it with master. However, I think we can link your tutorial in /docs

Let me know what you think.

Issue #93, opened 2013-04-05
deal with scroll panes


Platane:
Hello,

I appreciate a lot your work, I implement it in a web application project and have some issues, I thought it could be useful to share it.

Our application got a lot of panel that have toolbars. I wanted to make a step with an element  inside a scrollable panel, but it failed. That is what I have fixed here.

#### _getOffset method lack consideration for scrollpane

Firstly, there is a mistake with the _getOffset method, because it doesn' t consider scroll panel. Indeed if you want it to work with them, you have to substrat the scrollx and scrolly value, because scrolling make the element inside the scrollpane go up relatively to the window viewport.
I didn't try to fix it, instead I use jQuery which does the work pretty well.

#### highlight an element in a scrollpane

Secondly if the system is able to scroll the window screen to make the element visible, it can't scroll other panel.

I write a method which adjust the scrolling of all the parent of the element to make it visible. Again this function use jQuery.

I assume you choose to not work with jQuery, so my contribution is not so much helpfull, but  you can get inspired if you want to solve this minor issues.

the source code of my fixes and a example are visible [here](https://github.com/Platane/intro.js)

thanks you!

pavankumarkatakam:
Hi, the combination is very nice, i want to implement introjs in different div containers it is helpfull to keep scrollmeVisible in to intro js, well im thinking to create a merge request.hey @afshinm guide me to do so,

PR #388, opened 2015-01-06
Fixes offset calculation for scrolled elements

Anyqax:
fixes Issue #93 
see updated answer @https://stackoverflow.com/questions/442404/retrieve-the-position-x-y-of-an-html-element/442474#442474


bgruening:
:+1: this looks also good. @afshinm what do you think?

rjmorris:
I found that this messes up the positioning in Chrome, whether the page has been scrolled or not.

afshinm:
this should be fixed.

Issue #381, opened 2014-12-11
Stage number displaced


Bonsanto:
I am using **bootstrap 3.3.1** and **intro.js 1.0.0** and the number of the stage that shows the element focused by the intro gets displaced when the **bootstrap.min.css** and the intro **introjs-nassim.css** are included in the same project. Do you have any hint, you can see the problem in the next pictures.

With both styles.

<p align="center">
<img src="https://dl.dropboxusercontent.com/u/6280514/ShareX/2014/12/2014-12-11_15-46-55.png"></img>
</p>

Without the bootstrap style.

<p align="center">
<img src="https://dl.dropboxusercontent.com/u/6280514/ShareX/2014/12/2014-12-11_15-51-24.png" align="center"></img>
</p>

benthemonkey:
Duplicate of #361

Another related PR: #468

PR #393, opened 2015-01-11
Fix stage number positioning with bootstrap 3

dziemian007:
This is fix for using intro.js with bootstrap 3 and having stage number misplaced, issue has been described there[https://github.com/usablica/intro.js/issues/381]. (thanks to @Bonsanto)

This problem affects all layouts (not only introjs-nassim.css, as issue mentioned above may suggest).

Basically it's caused by bootstrap 3 policy for `box-sizing: border-box` everywhere. I've added `content-box` to `*`, and not only `introjs-helperNumberLayer`, because it affects more elements (`introjs-tooltip` f.ex.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,7 @@ Want to learn faster and easier? Here we have **Instant IntroJs**, Packt Publish
    - Fix bad sizing with Bootstrap 3
    - Add disable interaction ability
    - Fix code styling issues and many minor bug fixes
+   - Fix stage number positioning with bootstrap
 
  * **v0.9.0** - 2014-05-23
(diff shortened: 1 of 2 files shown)


veggiemonk:
:+1:

Bonsanto:
Wow I solved it modifying some .CSS but it was more like a temporal solution

DanielApt:
+1 Would love to see this added to the release soon.

Is anybody maintaining this repo anymore?

BeWilled:
Did not work for me. Tooltip blocked the highlighted section

PR #395, opened 2015-01-21
correct top position on firefox

mhidou:
Hi,

I've noticed that the top position don't work well on firefox on ubuntu. So I changed element.offsetTop to element.getBoundingClientRect().top .

https://developer.mozilla.org/en-US/docs/Web/API/element.getBoundingClientRect

I've tested on firefox and chromium and it's ok.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -1108,17 +1108,17 @@
     elementPosition.height = element.offsetHeight;
 
     //calculate element top and left
-    var _x = 0;
+    /*var _x = 0;
     var _y = 0;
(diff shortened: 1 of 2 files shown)


intro.js
```diff
@@ -1108,17 +1108,17 @@
     elementPosition.height = element.offsetHeight;
 
     //calculate element top and left
-    var _x = 0;
+    /*var _x = 0;
```
mysterlune: @mhidou, if the code is unused, please remove.

bgruening:
:+1: The dead code can be removed after merge if @mhidou has no time to work on it.

PR #407, opened 2015-02-19
Workaround for an infinite recursion bug

Chris-Turnbull:
Hi there

In version 0.6 you could pass a jQuery element into the steps array. This still works fine in version 1... except the new _cloneObject() function goes into an infinite recursion loop if you do this.

I've written a workaround which seems to work fine and fix the problem for us.


intro.js
```diff
@@ -262,7 +262,11 @@
       }
       var temp = {};
       for (var key in object) {
-          temp[key] = _cloneObject(object[key]);
+          if (object[key] instanceof jQuery) {
```
afshinm: Here you need to check if the jQuery is defined. I think this line can throw an error.

Chris-Turnbull: Oh, my apologies for that, we always include jQuery on our pages so I didn't notice.

afshinm: Cool!

afshinm:
merged.

PR #410, opened 2015-02-25
Added a close "X" to the help layer

mindblender:
Added an "X" to the top right of the help layer as another way to dismiss the help tips.


afshinm:
Although this is a cool feature but we have a lot of ways to exit the intro. Thank you for your help :)

PR #411, opened 2015-02-26
Prevent tabbing behind overlay

mindblender:
This change prevents a user from focusing on elements behind the overlay.  It checks to see if the focused item  is within the helplayer container and if it is not, then it sets focus to the 'next' button.


afshinm:
Can you please give me a demo link?

PR #412, opened 2015-02-26
The onExitCallback was not getting called when pressing Esc or clicking ...

mindblender:
...on the overlayLayer.  This was because the _exitIntro method call was being executed before the check for the introExitCallback.

Also added a check to see if the intro has completed when pressing enter on the skip button.


afshinm:
This is great. Thank you @mindblender

PR #415, opened 2015-03-15
Fallback to document if element null in _getOffset

dperish:
Necessary to support steps with no element in project using KendoUI and Bootstrap 2 (CleanZone template)

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -1101,6 +1101,9 @@
   function _getOffset(element) {
     var elementPosition = {};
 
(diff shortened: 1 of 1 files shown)


intro.js
```diff
@@ -1101,6 +1101,9 @@
   function _getOffset(element) {
     var elementPosition = {};
 
+    // Use document as element if not supplied
+    element = element || document;
```
afshinm: Why `element` should be empty?

PR #417, opened 2015-03-17
add ability to skip the step if the given element does not exist

ratiw:
Sometimes, I have elements hidden (or remove from DOM) and I don't want IntroJS to display it as floating tip.

With little modification, if the element defined in the step does not exist, it will just skip the step without displaying it as floating tip. But if the element is present, it will display that intro step.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -89,11 +89,17 @@
         var currentItem = _cloneObject(this._options.steps[i]);
         //set the step
         currentItem.step = introItems.length + 1;
+        var givenElementNotFound = false;
         //use querySelector function only when developer used CSS selector
         if (typeof(currentItem.element) === 'string') {
           //grab the element with given selector from the page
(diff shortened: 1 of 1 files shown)


kadishmal:
:+1: 

Here is how I done this:

        var Guides = [
		{
			element: '.item-creator-select:visible',
			intro: 'Some text'
		},
		...
        ];

	new introJs().setOptions({
		tooltipPosition: 'left',
		steps: _.filter(Guides, function (element) {
			return $el.find(element.element).length;
		})
	}).goToStep(parseInt(e.target.dataset.goto, 10) || 1).start();

Notice the `:visible` selector.

afshinm:
This is a good change but it doesn't solve the problem when the target element is `display: none;`.

afshinm:
For addressing this one, @kadishmal solution seems good. We shouldn't change that part. Thank you @ratiw

PR #421, opened 2015-03-21
override global tooltipPosition = "auto"

eballeste:
This update allows you to use the "auto" tooltipPosition value on a global level, but override it on a step by step basis.


afshinm:
Can you please explain it more?

eballeste:
Well, it basically let's you set to tooltip = auto setting on a global level but then override it on each step so basically if you want all of them to be set to auto you can do it on a global level and then specify a position by overriding it for one particular step.

PR #427, opened 2015-05-07
Adding Helper Layer User Specific Adjustment Options

anujkumartech:
At times we have observed that helper layer doesn't align with position relative selected elements. This two options leftAdjustLightBox and topAdjustLightBox ( which are to be provided along with step option) will update the current steps light box position accordingly.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -633,11 +633,23 @@
         widthHeightPadding = 0;
       }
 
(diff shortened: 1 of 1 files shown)


thotmx:
@anujkumartech, do you have an example to know how to use it?

Thanks in advanced.

Issue #369, opened 2014-11-08
Step number has wrong position with auto positioned 'floating' tooltip


yuvale:
Here's http://usablica.github.io/intro.js/example/auto-position/index.html with a thin browser window:

![image](https://cloud.githubusercontent.com/assets/2012892/4964254/bebf7644-675b-11e4-851e-67811f90d22d.png)

The step number isn't showing - it's actually positioned off the window, to the left of the underlying text.

I think I found the root cause:

The 'floating' position type was added in be96db02967e9144695e7e9b99ceaa1a11718b4f, where it can only be used when there's no target element. Since there's no visible element on which to put the number, this code was added to align the number layer with the tooltip layer in case of 'floating' position:

    if (typeof(helperNumberLayer) != 'undefined' && helperNumberLayer != null) {
        helperNumberLayer.style.left = '-' + ((tooltipOffset.width / 2) + 18) + 'px';
        helperNumberLayer.style.top  = '-' + ((tooltipOffset.height / 2) + 18) + 'px';
    }

Since pull requst #273 however, the 'floating' position can also be used *with* a target element if 'auto' positioning is used (i.e. _determineAutoPosition can return 'floating'). In this case the code above no longer makes sense, since it moves the number layer away from the target element.

I think the way to solve this is to only run the above code if there's no target element, as initially intended. If there is a target element, the step number is already positioned correctly.

PR #368, opened 2014-11-08
Fix issue 356 - no interaction with the target element.

yuvale:
An attempt to fix issue #356.
See the commit message for details.

PR #437, opened 2015-06-18
disableInteraction bugfix

hoschnok:
the disableInteraction seems always to be true. turned out there is an issue with the z-indexes. tooltipReferenceLayer has to be lower than helperLayer for it to work. disableInteraction option works correctly after this little fix.

### DIFF
--- a/introjs.css
+++ b/introjs.css
@@ -65,7 +65,7 @@ tr.introjs-showElement > th {
 
 .introjs-tooltipReferenceLayer {
   position: absolute;
-  z-index: 10000000;
+  z-index: 9999997;
   background-color: transparent;
   -webkit-transition: all 0.3s ease-out;
      -moz-transition: all 0.3s ease-out;


afshinm:
Thank you. Can you please give me demo link?

yuvale:
Note that if the tooltip is auto-positioned on top of the target element, then this fix will cause the target element to hide the tooltip.
Here's the example from #369 with this pull request applied:
image

hoschnok:
I did not know the tooltip can be positioned on top of the interactive area. I will probably have a look at this again.

yuvale:
Actually I created pull request #368 for this same issue a while ago, with a different approach.

afshinm:
@yuvale I merged your PR. Do we need this change now?

yuvale:
No, both PRs solve the same issue.

afshinm:
Thank you @hoschnok @yuvale

PR #439, opened 2015-06-22
Hide Next button on last step

Maado:
Hide Next button on last step


afshinm:
Thank you for this change. We don't need to hide the next button, you can control in using css styles instead of JavaScript.

Issue #445, opened 2015-07-20
Enter key ends tour for multipage intro, instead of progressing to next page


ericenns:
Enter key goes to next step for every item but when no steps are left, it ends the tour instead of progressing to next page.

Arrow keys work for this, but I recently had a accessibility compliance test done of my site and they brought up that the enter key should progress to the next page.

ericenns:
fixed in 6fd0378313c5b55c737361d9c704ec081d5b65f9

PR #446, opened 2015-07-20
Enter key will now progress to next page for multi-page intro

ericenns:
Addresses issue #445

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -214,7 +214,11 @@
             _previousStep.call(self);
           } else if (target && target.className.indexOf('introjs-skipbutton') > 0) {
             //user hit enter while focusing on skip button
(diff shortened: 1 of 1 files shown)


afshinm:
I think we should modify multipage example because this behavior is correct for single page intros, right?

ericenns:
Yes so basically before, enter key would progress to next step until it hit the last where it would end the tour. In a single page tour it will still just end the tour as normal.

afshinm:
Right, so let's change the multi-page example instead of the JS code. Can you help me? :)

ericenns:
fixed in 6fd0378313c5b55c737361d9c704ec081d5b65f9

afshinm:
Thank you @ericenns

PR #448, opened 2015-07-23
only apply .introjs-arrow.left if the tooltip was not re-positioned

traviswingo:
When the tooltip falls below the window, we reposition it to push up
into the existing viewport. This fix also adds in the proper arrow
styling so that it falls on the bottom of this tooltip and points to
the actual content.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -473,8 +473,9 @@
           // Modify so that the bottom of the tooltip connects with the target
           arrowLayer.className = "introjs-arrow left-bottom";
           tooltipLayer.style.top = "-" + (tooltipHeight - targetOffset.height - 20) + "px"
+        } else {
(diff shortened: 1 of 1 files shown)


afshinm:
Can you please give me a demo of this bug?

kleinph:
This should show a valid demo of this bug: https://jsfiddle.net/0x20urxg/

kleinph:
But you can also see it easily in the code as you set `arrowLayer.className` in an `if` statement and then reset it to another value anyway. It is different in the "left" case of the switch as you can see a few lines below.

traviswingo:
Thanks for that, Philipp. Been too busy to come back and provide a demo.

afshinm:
Thank you guys.

PR #450, opened 2015-08-05
Added an extra check to determine how to load the module

### LINKED ISSUE #449: Insufficient check for existence of Common.js
I found out that this check

```
if (typeof exports === 'object') {
    factory(exports);
}
```

is insufficient by itself, to determine whether Commonjs is used.

I am already using [CSSlint](https://github.com/CSSLint/csslint) which [creates](https://github.com/CSSLint/csslint/blob/master/release/csslint.js#L25)  a global `export` object without Commonjs actually being in use.

As a result Intro.js tries to load as a Common.js module instead of Require.js module, which I am using

mkarajohn:
resolves #449

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -7,7 +7,7 @@
  */
 
 (function (root, factory) {
-  if (typeof exports === 'object') {
+  if (typeof exports === 'object' && typeof module == 'object') {
     // CommonJS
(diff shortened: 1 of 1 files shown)


mkarajohn:
disregard actually.

PR #455, opened 2015-08-25
Fixes #454: Position tooltip layer and its arrow more intelligently

### LINKED ISSUE #454: Position tooltip layer and its arrow more intelligently
When top/bottom positioning is used and the target element is on the very left or right of the page, the tooltip goes beyond the visible area to the left or right making its content unreadable.

The same happens when left/right positioning is used and the target element is on the very top or bottom of the page, the tooltip goes beyond the visible area to the top or bottom making its content unreadable.

This PR fixes these issues by positioning the tooltip more intelligently. It checks if the tooltip would go beyond the window size (left, right, top, bottom) and sets the values accordingly.

kadishmal:
Fixes #454.


afshinm:
Can you please update this PR with upstream?

kadishmal:
@afshinm done!

PR #458, opened 2015-09-01
Fixes a case where .oncomplete and .onexit callbacks run both

mkarajohn:
Fixes .onexit callbacks firing when user hits Enter on 'Done' button, of last step.

Since there is a clear distinction between a 'complete' and an 'exit' event, I think it is wrong behaviour to fire 'exit' callbacks when the intro was 'completed'

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -211,13 +211,13 @@
             //user hit enter while focusing on previous button
             _previousStep.call(self);
           } else if (target && target.className.indexOf('introjs-skipbutton') > 0) {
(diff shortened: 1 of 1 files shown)


mkarajohn:
@afshinm is this acknowledged?

rhyous:
I disagree. Complete is when done is clicked. Exit is whenever IntroJs closes, which includes after Complete runs. 

So when clicking done, two events should occur: oncomplete and onexit.

PR #490, opened 2015-11-18
Add a way to prevent transitioning to a next step

evach-vf:
Here’s the context.  Our application is passing all multi-pages steps at once in JSON .
We handle the redirection using `onbeforechange` callback and we resume the intro by calling the `goToStep` method.

Currently, when the intro redirects to another page, it starts to transition to the next step and gets interrupted half-way by the redirection.  This causes the step to be seen twice.
It works but it flickers!

With this patch, by calling `introJs.disableAnimation(true)`, it will prevent the transition to the next step and the redirection completes without flickering.
The result is a smoother transition between pages.


bozdoz:
I don't believe this PR is complete.  Also this is quite out-dated.  My apologies of course for the extremely late response to this.  Cheers, and thanks!

PR #500, opened 2015-12-17
Hide Elements + Set Right step number

lexfernandez:
Fix Display only visible element + set right step number


bozdoz:
Sorry for the extremely late reply.  :smile: Could you explain what this PR does?

PR #514, opened 2016-01-18
Option: Set padding around helper element

BrianLaughlan:
Having a fixed value for the additional padding around the helper element is not flexible. This small change will make it configurable.


pixelize:
@afshinm any merge incoming ? :)

afshinm:
@BrianLaughlan and @pixelize I'm so sorry for this delay. thanks!

PR #517, opened 2016-01-26
Added callback for onskip event (clicking skip inside tooltip container)

labithiotis:
Have added an event to invoke a onskip event similar to oncomplete. This is different to generic on exit event which is invoked by hitting ESC or clicking overlay. This will only be invoked when a user hits the skip button inside the container, allowing you to track when user intentionally wants to skip intro and thus know to saving that action. The onexit is still invoked so existing configurations will still work in same manner.


Geczy:
Exactly what I needed!

afshinm:
hey @labithiotis, thanks for this. could you please rebase it though?

labithiotis:
@afshinm Resolved conflict.

PR #518, opened 2016-01-27
Make hint pulse animation work on ie10

BrianLaughlan:
The keyframes are only set for -moz and -webkit .. there is no standard keyframes set, this means that the pulse on the hints will not work on ie10

There was also a typo on the moz-keyframes ... it said: "intrjspulse" instead of "introjspulse"

### DIFF
--- a/introjs.css
+++ b/introjs.css
@@ -406,48 +406,71 @@ tr.introjs-showElement > th {
   opacity: 0;
 }
 
-@-moz-keyframes intrjspulse {
- 0% {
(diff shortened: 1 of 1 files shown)


zarvox:
Heh.  I found the same issue and wrote ~the same patch, then searched the PRs and found this. :)

Would love to hear if there's likely to be an update soon, or if I should just vendor this bugfix into our repo!

afshinm:
Nice one, could you please rebase with upstream so I can merge? @BrianLaughlan

PR #521, opened 2016-02-04
tooltipPosition option is not applied, when steps are defined in code

gerwald:
If tooltipPosition option is set, but the steps are added via setOptions (and no explicit option is defined), the tooltipOption is not applied

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -111,6 +111,10 @@
 
           currentItem.element  = floatingElementQuery;
           currentItem.position = 'floating';
+        } else {
+            // using default options, if not set
+            if (!currentItem.position)
+                currentItem.position = this._options.tooltipPosition;
         }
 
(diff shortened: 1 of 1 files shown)


bozdoz:
Not sure what kind of PR this is.  Is this resolving a bug?

gerwald:
yes it solves a bug - like I described in the commit message:
If tooltipPosition option is set, but the steps are added via setOptions (and no explicit option is defined), the tooltipOption is not applied

bozdoz:
Could you give an example of when this would occur?  Thanks! :smile:

gerwald:
Sorry, that was two years ago, no idea...

bozdoz:
Thanks for the help.  Going to have to close this, as I am unaware of what it does.  Sorry about that, and for the lateness of this reply.

PR #534, opened 2016-02-17
Add new template 'flattener'

711fumi:
I tried to create a new template.
Please merge if could use.

image


afshinm:
This is great @711fumi!

PR #540, opened 2016-02-25
Changed version number to a valid semver value.

rjmorris:
Changed the version number from 2.0 to 2.0.0 to meet the semver requirements. Changed not just in package.json, but everywhere else it appears.


afshinm:
I think I already addressed this one. btw, can you please rebase this with upstream?

PR #541, opened 2016-02-25
Removed introjs-fixedTooltip class when moving to a step whose element isn't fixed.

rjmorris:
The introjs-fixedTooltip class is added to the tooltipReferenceLayer when a step's element is fixed. However, the class is never removed, so the tooltip positioning is messed up for later steps. This patch removes the introjs-fixedTooltip class (if any) when the step's element isn't fixed.


icai:
+1

afshinm:
thanks

PR #543, opened 2016-02-26
Add web components compatibility (polymer)

AlexBeauchemin:
Allow introjs to be used in web components (polymer) as web components root elements are not html tags (body, html...)


joeblew99:
@AlexBeauchemin 
Any idea why this is blocked ?

afshinm:
just added this for the next version.

afshinm:
added using #4a12cdf

PR #545, opened 2016-03-01
Minor updates/fixes to GH pages site for development/testing compatibility

tekwiz:
1. Removing http/s protocols from src & href URLs
2. Changing example links to be relative URLs


afshinm:
thanks @tekwiz

PR #556, opened 2016-03-13
Remove multiple overlay layers if present

ashwinaggarwal:
@afshinm, ideally there should be one _overlay_ layer present, but in cases when `introJs.exit()` is called and immediately `introJs().start()` is called, there as two _overlayLayer_ present in dom, so when `introJs.exit()` is called again, one _overlayLayer_ is removed and other tends to exist unless user clicks on it to remove it.

This fix will remove all the _overlayLayers_ present.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -356,21 +356,21 @@
    * @param {Object} targetElement
    */
   function _exitIntro(targetElement) {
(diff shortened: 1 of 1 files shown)


afshinm:
nice one

PR #558, opened 2016-03-15
Remove `introjs-fixedTooltip` if not needed

Kmaschta:
Let's consider we got both a fixed and a not fixed elements (order matters) handled by intro.js on the same page. On Firefox, intro.js would first add the `introjs-fixedTooltip` on the helper layer. That's normal, as the first element is fixed. Yet, it won't remove this class on the second one, which is not fixed.

This PR fixes the issue.


icai:
+1

afshinm:
thanks guys

PR #572, opened 2016-04-20
Fix blocking error when a tooltipbutton doesn't exist + scroll improvements

JvJefke:
Hi,

An Error occurs when there is a guide with only 1 step. There is no prev- or next tooltip button. The code errors on that and the guide doesn't show the step properly. This pull request prevents the occurence of the error so that the function can continue and the guide can be shown correctly.

I also improved the scrolling. Now the scrolling takes both the tooltip ant the target into account for scrolling. The scrolling also always happens after the tooltip has ben set. This fixes som targeting offset problems.


afshinm:
could you please rebase?

Issue #579, opened 2016-05-14
Hide next button when there is no next step.


DeliRadioAdmin:
We can see next button even if there is no where for user to go.

We should be allowed to configure buttons on a step in some capacity

HarrisonOg:
@afshinm any plans on getting this in?

afshinm:
I have added the option in the previous version.

PR #580, opened 2016-05-19
Change for Hide next button when there is no next step. #579

anujkumartech:
To address this issue new parameter in option values added(named it as 'hideLabelOnExit')

hideLabelOnExit = 'next' will hide next button
hideLabelOnExit = 'previous' will hide back button
hideLabelOnExit = 'both' will hide both of them


vas1468:
Can we please merge this branch?

afshinm:
we have this option now

Issue #564, opened 2016-03-25
Urgent Problem with hints in SPA


babnik63:
When changing routes in SPA's hints stays on the window but every thing is changed.
I think having somthing like: introjs().removeHints() could solve the issue.

babnik63:
I did some tricks and failed, I think it needs a public function.

afshinm:
Can you please prepare an online demo for this issue? a jsfiddle maybe

babnik63:
I'm using Meteor so jsfiddle is not possible. I need to be able to remove hits manually by code, then I can use this function in my routing changes.

tonyspb:
@afshinm 
https://jsfiddle.net/0xbxmvu2/2/

the thing is one can't hide unused hints and show new hints.

babnik63:
@tonyspb Yes this is exactly my problem.

afshinm:
ah I see. will apply a new patch today/tom and release v2.0.1.

afshinm:
guys @babnik63 @tonyspb, I've added a refresh function for hints that re-arranges all hints. 
Do you think this can help you?

babnik63:
Hi,
I think we can call this Refresh while changing routes or any thing else.
BUT: the refresh function should not do anything if the hint is not active and called before. the refresh function is not the showhint() function

babnik63:
I saw you changes
I have tested     introJs().refresh(); but nothing changed. the hint element is no longer on page but the circles are showing.
I think we need to remove old hints so a removeHints function is a better solution.
is there any?

babnik63:
@afshinm what is you idea about the above comment?

afshinm:
hey @babnik63. I added "hideHint" function to the public api last night. could you please check this and lmk if it is the one you are looking for?

babnik63:
Dear Afshin
hideHint() needs stepid, but I need something to call and remove all active hints.
this is my suggested code:
  ` function _hideAllHints() {
        _removeHintTooltip.call(this);
        var hint = this._targetElement.querySelectorAll('.introjs-hint');

        if (hint) {
            var i;
            for (i = 0; i < hint.length; i++) {
                hint[i].className += ' introjs-hidehint';
            }

        }

       
    };`
Thanks

afshinm:
thanks @babnik63. will add this for v2.2.0.

afshinm:
done @babnik63

PR #588, opened 2016-05-31
closes #587 add API call to hide all hints

### LINKED ISSUE #587: Add API call to hide all hints
As a developer I want to be able to show and hide all hints with one function call, instead of having to iterate over all steps and calling `hideHint(step)` for every step.

makkes:
- added public API function
- added a button in the hints example to hide all hints


makkes:
Closed with fix for #564

afshinm:
ah sorry and thanks @makkes.

PR #589, opened 2016-06-03
introjs.css bugfix

aramix:
removed quotes around the keyframe's name
fixes a typo


afshinm:
already fixed, thanks.

PR #595, opened 2016-06-24
fixes #345 Add option to skip missing elements

### LINKED ISSUE #345: Do not show tooltip
how to not show  tooltip, if not find the id?

makkes:
- When defining a product tour using e.g. a JSON file containing all
  the steps and CSS selectors, one wants to be able to just skip those
  steps that target elements the user currently doesn't see, anyway.
  So I added an option 'skipAbsent' to do exactly this.


afshinm:
sure, cool. do you have any usecase for this?

makkes:
Actually yes. We want to have a single JSON file in our application that defines all the steps for a product tour. Since different users see different things (depending on their access rights inside of our application), some steps must be skipped (e.g. the step that explains administrative features). Intro.js should just skip these silently then.

ext0:
Likewise, great usecase. Currently implementing this feature in my own project, as some elements are not visible until future user progression.

PR #596, opened 2016-06-24
Added support for promises in the beforeChange event, allows the user…

ext0:
… to perform asynchronous tasks before continuing with the next element.

Allows the user to delay (but not stop) the progression (for asynchronous tasks that must be executed prior to progressing) to the next element via promises. In order to utilize, the onbeforechange callback can return a Promise object that is resolved when the user wants to continue with the progression.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -330,8 +330,16 @@
     }
 
     var nextStep = this._introItems[this._currentStep];
+    var beforeChangePromise = null;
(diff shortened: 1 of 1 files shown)


afshinm:
that's a good idea to have this but if we want to change this to promises, then we need to change all of the callbacks not just one.

swifty-team:
what the status of this?

vitorleonel:
what the status of this?

PR #606, opened 2016-06-30
Intro.js with enhancements

noammohr:
1) In order to create a custom progress bar, the logic concerning the built-in progress bar was moved into the `_createProgressBar()` and `_updateProgressBar()` functions, a parameter `monkeypatch` was added to `introJs()` and `IntroJS()` and stored in `this._monkeypatch`, and those newly created functions check `this._monkeypatch` for an attribute with the same function name to run that function instead of the built-in one if provided. That is, passing in a parameter `monkeypatch = {'_createProgressBar': function(){...}, '_updateProgressBar': function(){...}}` will substitute these functions for the built-in ones and display the custom progress bar. (Note that it may make sense, both for monkeypatching and for general readability, to do the same for other parts of the tour, for example, `_createBullets()`, `_updateBullets()`, `_createButtons()`, `_updateButtons()`, etc. Also, it may make sense to find a better way to monkeypatch, for example, by moving `IntroJs` into the global scope.)

2) Several new options were added, including:
- `doneClass`: a class for the Done button to make it display with different style than the next/back/skip buttons. Creating a separate button rather than reusing the skip button is probably a better solution.
- `doneClassInsteadOfDefault`: This is a hack used to get `doneClass` to replace `introjs-button` instead of making changes on top of that default class. Losing the `introjs-button` class breaks anything that permits going back to earlier steps in the tour, like the Back button and the Bullets.
- `showBackButton`: Setting this to false removes the Back (Prev) button, so users can only go forward in the tour.
- `padding`: intro.js creates a 5-pixel white border around a highlighted element. This option allows the size of the border to be customized.
- `headerHeight`: Allows setting the size of an area at the top of the screen which is considered off-screen for the purpose of displaying the tour. This can ensure that any page masthead does not get covered by the tour.

3) Added new public function `getStep()` to return the step of the tour we're currently on.

4) I created `_getCallback()` and `_setCallback()` functions for events to (a) avoid repeated code, and (b) to allow the user to add functionality to an event that already has a function, rather than replacing the already existing function (which is the only possibility currently). I set it so that adding functionality is the default, and setting the `overwrite` parameter to true replaces the existing functionality. This let's me create a way for all tours on my site to work, and then allow individual tours on this page or that page to add on their own thing.

5) Some minor changes to flow and naming for clarity.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -11,6 +11,7 @@
  *   skipLabel
  *   doneLabel
  *   doneClass
+ *   doneClassInsteadOfDefault
  *   tooltipPosition
  *   tooltipClass
  *   highlightClass
@@ -89,8 +90,10 @@
       skipLabel: 'Skip',
       /* Done button label in tooltip box */
       doneLabel: 'Done',
-      /* Default tooltip box position */
+      /* Class to add on Done button */
       doneClass: null,
+      /* Should doneClass replace (rather than add to) default button class? Does not work well with showBackButton or showBullets. */
+      doneClassInsteadOfDefault: null,
       /* Default tooltip box position */
       tooltipPosition: 'bottom',
       /* Next CSS class for tooltip boxes */
@@ -133,6 +136,10 @@
       /* Height of header area at the top not to be covered by tooltip. */
       headerHeight: 0
     };
+    if (this._options.doneClassInsteadOfDefault === null) {
+      // Options that allow going backwards on a tour require keeping the default introjs-button class.
+      this._options.doneClassInsteadOfDefault = !(this._options.showButtons || this._options.showBullets);
+    }
   }
 
   /**
@@ -1082,7 +1089,11 @@
     } else if (this._introItems.length - 1 == this._currentStep || this._introItems.length == 1) {
       skipTooltipButton.innerHTML = this._options.doneLabel;
       if (this._options.doneClass != null) {
-        skipTooltipButton.className = 'introjs-button ' + this._options.doneClass;
+        if (this._options.doneClassInsteadOfDefault) {
+          skipTooltipButton.className = this._options.doneClass;
+        } else {
+          skipTooltipButton.className = 'introjs-button ' + this._options.doneClass;
+        }
       }
 
       if (this._options.showBackButton) {
--- a/intro.js
+++ b/intro.js
@@ -3,6 +3,56 @@
  * https://github.com/usablica/intro.js
  *
  * Copyright (C) 2016 Afshin Mehrabani (@afshinmeh)
+ *
+ * Options:
+ *   steps
+ *   nextLabel
+ *   prevLabel
+ *   skipLabel
+ *   doneLabel
+ *   doneClass
+ *   tooltipPosition
+ *   tooltipClass
+ *   highlightClass
+ *   exitOnEsc
+ *   exitOnOverlayClick
+ *   showStepNumbers
+ *   keyboardNavigation
+ *   showButtons
+ *   showBackButton
+ *   showBullets
+ *   showHelperLayer
+ *   showProgress
+ *   scrollToElement
+ *   overlayOpacity
+ *   padding
+ *   positionPrecedence
+ *   disableInteraction
+ *   hintPosition
+ *   hintButtonLabel
+ *   headerHeight
+ *
+ * Methods:
+ *   clone()
+ *   setOption(option, value)
+ *   setOptions(options)
+ *   start()
+ *   getStep()
+ *   goToStep(step)
+ *   nextStep()
+ *   previousStep()
+ *   exit()
+ *   refresh()
+ *   onbeforechange(providedCallback)
+ *   onchange(providedCallback)
+ *   onafterchange(providedCallback)
+ *   oncomplete(providedCallback)
+ *   onexit(providedCallback)
+ *   onhintsadded(providedCallback)
+ *   onhintclick(providedCallback)
+ *   onhintclose(providedCallback)
+ *   addHints()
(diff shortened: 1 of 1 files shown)


noammohr:
@afshinm This is up to date with master.

afshinm:
thanks @noammohr. Can you please explain your changes? or maybe an online demo.

I'm happy to find a better solution for your changes and avoid using `monkeyPatching`.

afshinm:
any updates on this @noammohr? Have you published your changes?

noammohr:
@afshinm Changes have not yet been published.

1) In order to create a custom progress bar, the logic concerning the built-in progress bar was moved into the _createProgressBar() and _updateProgressBar() functions, a parameter `monkeypatch` was added to introJs() and IntroJS() and stored in `this._monkeypatch`, and those newly created functions check `this._monkeypatch` for an attribute with the same function name to run that function instead of the built-in one if provided. That is, passing in a parameter `monkeypatch = {'_createProgressBar': function(){...}, '_updateProgressBar':function(){...}}` will substitute these functions for the built-in ones and display the custom progress bar.

2) Several new options were added, including:
  * `doneClass`: a class for the Done button to make it display with different style than the next/back/skip buttons. Creating a separate button rather than reusing the skip button is probably a better solution.
  * `doneClassInsteadOfDefault`: This is a hack used to get doneClass to replace `introjs-button` instead of making changes to that default class. Losing the `introjs-button` class breaks anything that permits going back to earlier steps in the tour, like the Back button and the Bullets.
  * `showBackButton`: Setting this to false removes the Back button, so users can only go forward in the tour.
  * `padding`: intro.js creates a 5-pixel write border around a highlighted element. This option allows the size of the border to be customized.
  * `headerHeight`: Allows setting the size of an area at the top of the screen which is considered off-screen for the purpose of displaying the tour. This can ensure that any masthead does not get covered by the tour.

3) Added new public function getStep() to return the step of the tour we're currently on.

4) I created _getCallback() and _setCallback() functions for events to (a) avoid repeated code, and (b) to allow the user to add functionality to an event that already has a function, rather than replacing the already existing function (which is the only possibility currently). I set it so that adding functionality is the default, and setting the `overwrite` parameter to true replaces the existing functionality. This let's me create a way for all tours on my site to work, and then allow individual tours on this page or that page to add on their own thing.

noammohr:
@afshinm I wanted to check in with you again about these changes. They have been published.

noammohr:
@afshinm I wanted to check back on this. Have you had a chance to take a look?

afshinm:
@noammohr yeah thanks for this Noam. Will have a look this week, sorry I was busy.

noammohr:
@afshinm Just checking back again. Thanks.

noammohr:
@afshinm Can you give me some sense of when to expect a response? I don't want to press the point too hard, but we did pay for the maximum amount of support.

afshinm:
hey @noammohr, please accept my apologies for the delay on this. I will have a look this weekend and will comment the issues on this PR. 

Thanks!

intro.js
```diff
@@ -3,6 +3,58 @@
...
+ *   onbeforechange(providedCallback)
+ *   onchange(providedCallback)
+ *   onafterchange(providedCallback)
+ *   oncomplete(providedCallback)
+ *   onexit(providedCallback)
+ *   onhintsadded(providedCallback)
+ *   onhintclick(providedCallback)
+ *   onhintclose(providedCallback)
+ *   addHints()
+ *   hideHint(stepId)
```
afshinm: Nice to have methods here.

intro.js
```diff
@@ -1605,22 +1726,80 @@
...
   };
 
-  var introJs = function (targetElm) {
+  /**
+   * Get the private callback function name from the public event function name.
+   *
+   * @param {String} functionName (e.g., 'onbeforechange')
+   * @returns {String} (e.g., '_introOnBeforeChangeCallback')
+   */
+  function _getCallback(functionName) {
```
afshinm: Nice one.

intro.js
```diff
@@ -498,15 +565,15 @@
         }
 
         _checkRight(targetOffset, tooltipLayerStyleLeft, tooltipOffset, windowSize, tooltipLayer);
-        tooltipLayer.style.bottom = (targetOffset.height +  20) + 'px';
+        tooltipLayer.style.bottom = (targetOffset.height +  15 + this._options.padding) + 'px';
```
afshinm: Why there is a 15px by default?

intro.js
```diff
@@ -702,7 +769,7 @@
 
       var currentElement  = this._introItems[this._currentStep],
           elementPosition = _getOffset(currentElement.element),
-          widthHeightPadding = 10;
+          widthHeightPadding = this._options.padding * 2;
```
afshinm: maybe same as the previous one, why there is a *2?

afshinm:
hey @noammohr, these changes seem great, thank for this.

Can you please make two different PRs so we can manage them better? Please move the progress bar change and monkey patching to another PR and keep the options changes, comment changes and add/get callback here in this PR.

afshinm:
kind reminder @noammohr

noammohr:
I have not completely forgotten, but it will take effort to find the time to separate out the monkey patching. I'll note, though, that `DoneClassInsteadOfDefault` doesn't play nice with every other option and therefore may not be ready for prime time. Also, I mentioned that currently you can overwrite function but can't add to current functionality, and when I changed it I set it so that adding functionality is the default -- really the default should be to overwrite for backwards compatibility. So these are also adjustments I'll make when I get to this.

PR #611, opened 2016-07-04
clarify license, wording and SPDX

graingert:
The MIT license is not viral copyleft as the current wording seems to imply. Either update the wording as in this PR or switch to something like AGPL 3+.


afshinm:
will merge this for the next version. thanks @graingert

Issue #466, opened 2015-09-17
issue in multipage intro example


manojverma086:
![Uploading Screenshot from 2015-09-17 20:12:17.png…]()
I tried it and when introduction finishes on last page then  there is an error:-  "uncaught TypeError: Cannot set property 'className' of null".

gsivaprabu:
I have comment below line is working now
![comment](https://cloud.githubusercontent.com/assets/2444608/14709555/d782c12e-07ee-11e6-8f9b-9f43390ce0f4.png)


` //oldReferenceLayer.querySelector('.introjs-bullets li > a.active').className = '';`

PR #619, opened 2016-07-15
Fixed issue when property wasn't able to be set, when showBullets was…

### LINKED ISSUE #434: Multi-page demo is throwing error
This isn't an issue with the code (I don't think, anyway), but the example for multi-page is broken.

It goes to the second page just fine, and then after hitting Next once, it throws:

```
TypeError: Cannot set property 'className' of null
http://usablica.github.io/intro.js/intro.js:745
```

Giegling:
… disabled

This commit fixes this error: “uncaught TypeError: Cannot set property
'className' of null”

Fixes #434 and #466

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -839,9 +839,10 @@
         _placeTooltip.call(self, targetElement.element, oldtooltipContainer, oldArrowLayer, oldHelperNumberLayer);
 
         //change active bullet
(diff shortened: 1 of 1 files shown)


afshinm:
thanks for this @Giegling

PR #620, opened 2016-07-26
Add hint count parameter to onhintsadded callback

makkes:
- This parameter denotes the count of hints that were actually
  rendered, which can actually be less than the configured hints, e.g.
  when the CSS selector doesn't select an element. Useful for having
  only one hint configuration for multiple views of single-page
  applications.


afshinm:
hey, thanks for this. It's possible to get this data inside the callback function.

PR #621, opened 2016-07-29
Remove double quote for keyframe name(Bugfix)

juliosueiras:
## Description:
Remove double quote for keyframe name(Bugfix) - Which resolve the issue of @keyframes missing name error(which also occur in meteor)

## Issue:
- `@keyframes missing name` error occur during meteor build with introjs in either bower or lib/ folder

## Solution:
- Removing double quotes of [line 443](https://github.com/usablica/intro.js/blob/master/introjs.css#L443)
 cause the error to disappear

### DIFF
--- a/introjs.css
+++ b/introjs.css
@@ -440,7 +440,7 @@ tr.introjs-showElement > th {
  }
 }
 
-@-webkit-keyframes "introjspulse" {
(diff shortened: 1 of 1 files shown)


afshinm:
thanks.

PR #623, opened 2016-08-11
Changed line-height to 13px to center the number on the indicator

leofontes:
Hey guys,

Really like introJS, but I didn't like how the numbers for the steps were not centered inside the circle and I believe this might be a common problem. 

To save everyone the trouble I changed the line-height for the span to be the same as the font-size.

Thank you, hope this helps.


afshinm:
well I can't reproduce this and i think the number is centered, can you please provide a screenshot?

PR #636, opened 2016-09-18
Previous Step support

olmec-akeru:
Added _previousStep internal variable

Added previousStep parameter to onbeforechange callback to provide access to the previous step for callback function.

Modified paramaters to pass introStep to callback and not introStep.element to make more information available to callback function

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -315,6 +315,7 @@
    */
   function _nextStep() {
     this._direction = 'forward';
+    this._previousStep = +this._currentStep;
 
     if (typeof (this._currentStep) === 'undefined') {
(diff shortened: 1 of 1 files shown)


olmec-akeru:
So you're not looking at pull requests?

afshinm:
sorry for the late response. No, I do review the PRs but I'm working on something else to add in parallel so it makes it a little bit difficult to review all PRs. thanks for this though and I will merge it later.

afshinm:
it is possible to get this data using `this` inside the callback function and passing this value is not necessary. thanks for this though.

PR #643, opened 2016-10-28
Add step

nblackburn:
Added convenience methods for adding single and multiple steps.


afshinm:
awesome. could you please update the wiki as well?

nblackburn:
@afshinm Updated wiki

afshinm:
awesome thanks

PR #649, opened 2016-11-09
Show hints

### LINKED ISSUE #616: Hiding all hints and then displaying them again
Hello, seems like hints don't work this way. For example I've got one button, which is "Show/Hide Hints". When hints aren't displayed, I call "introJs().addHints()" and it displays all hints. On the second click, I call "introJs().hideHints()", which removes all hints and after this I have a problem. When I click that button third time "introJs().onhitsadded()" is getting called, but no hints are shown.

nblackburn:
Fixes #616.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -1379,7 +1379,7 @@
         _showHint.call(this, hints[i].getAttribute('data-step'));
       }
     } else {
-      _addHints.call(this);
+      _populateHints.call(this, this._targetElement);
     }
   };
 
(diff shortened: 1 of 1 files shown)


Narretz:
@afshinm Would be cool if this could be merged. (the wiki needs to be updated, though).

nblackburn:
@Narretz Can't do that until this is merged.

afshinm:
Nice one! thanks for this

nblackburn:
Updated wiki

PR #657, opened 2016-11-23
Fixed issue when elements hidden by css are activated

GertjanVermeir:


afshinm:
thanks for this but we don't have any dependency to jquery or anything similar.

PR #665, opened 2016-12-20
Add new function `goToStepNumber`

doits:
refs #299

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,6 @@ Optionally, pass one parameter to `introJs` function to limit the presentation s
 
 <p align="center"><img src="https://raw.githubusercontent.com/usablica/intro.js/gh-pages/img/introjs-demo.png"></p>
 
-
 ## Documentation
 
 Please visit [Documentation](http://introjs.com/docs).
--- a/docs/docs/intro/api.md
+++ b/docs/docs/intro/api.md
@@ -67,6 +67,36 @@ introJs().goToStep(2).start(); //starts introduction from step 2
 
 -----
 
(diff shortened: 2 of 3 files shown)


afshinm:
cool thanks I will have a look

afshinm:
Sorry @nblackburn, which part of the PR?

nblackburn:
@afshinm Line 318

afshinm:
Ok, what is the difference between this and this one: 

```
    goToStep: function(step) {
      _goToStep.call(this, step);
      return this;
    },
```
https://github.com/usablica/intro.js/blob/master/intro.js#L1765

doits:
> Ok, what is the difference between this and this one: [...]

Not sure what you mean exactly, you mean the difference? See the commit msg:

> This allows to go to a specific step with the specific `data-step`
> attribute. This differs from `goToStep` in the way that `data-step`
> does not have be continuous to pick the desired element.
> 
> For example:
> 
> ```
> <div id="first" data-step='5'></div>
> <div id="second" data-step='9'></div>
> 
> // to get #second one can do now:
> introJs().goToStepNumber(9); --> #second
> 
> // or with goToStep:
> introJs().goToStep(2); --> #second
> ```

afshinm:
ah yeah right. sorry I didn't read it carefully (or I need to change my glasses).

Thanks for this, could you please update the docs as well so I can merge it today?

doits:
Added it, you have to correct `Available since` in ready though.

afshinm:
Perfect! thanks @doits 

thanks for review @nblackburn

Issue #667, opened 2016-12-21
Add yarn support


nblackburn:
Yarn is growing traction and it makes sense to add support for those that need it.

**Website**: https://yarnpkg.com

PR #670, opened 2016-12-21
Added yarn lock

nblackburn:
See #667


afshinm:
Sorry I'm not familiar with Yarn, so that file adds the Introjs to the Yarn inventory?

nblackburn:
@afshinm No, that is just the lock file needed. You will have to publish it via the publish command found in the CLI.

afshinm:
cool thanks

PR #675, opened 2017-01-11
Fixed going back one step after reaching the last step

radumpopescu:
Hi!
I found a bug in version 2.5.0.

If you go the the last step and then go back to previous step, the skip button can't be found, cause .introjs-skipbutton class has been overriden by this 9ae68362dd2177dcf9e7751586a51d17fc9b1688.
One solution I found is to search for the new class in case the first one is not found and also to change it back to the initial class.

I'm not very proud of that `if` so if you have a better option, do tell me!

Thanks!

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -811,7 +811,10 @@
           skipTooltipButton    = oldReferenceLayer.querySelector('.introjs-skipbutton'),
           prevTooltipButton    = oldReferenceLayer.querySelector('.introjs-prevbutton'),
           nextTooltipButton    = oldReferenceLayer.querySelector('.introjs-nextbutton');
-
+      if (!skipTooltipButton){
+        skipTooltipButton = oldReferenceLayer.querySelector('.introjs-donebutton')
+      }
+      
       //update or reset the helper highlight class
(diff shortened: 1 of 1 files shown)


afshinm:
Alright, I will have a look. thanks.

afshinm:
thanks for reporting this issue @radumpopescu 

I've used another approach to fix that:https://github.com/usablica/intro.js/commit/6e8a8cbb328fd021536171d3360dc84e5e6430c1

faisal-web:
**Hi i am facing the problem **
**i am using multipage intro my problems is when i click skip button stop intro but when i refresh my page intro is work my requirement is when i click skip button intro is stop using the code in below thankx  **

<script type="text/javascript">

      if (RegExp('multipage', 'gi').test(window.location.search)) {
//        introJs().start();
        introJs().setOption('doneLabel', 'Next page').start().oncomplete(function() {
          window.location.href = 'http://localhost/templated-ui/b7.php?multipage=true';
        });

        introJs().setOption("skipLabel").oncomplete(function () {
          window.location.href = 'http://localhost/templated-ui/b7.php';
        });

      }

    </script>

PR #676, opened 2017-01-16
Add available since tags for hints API docs

Simon-Briggs:
Hi, 

I got confused using the angular wrapper for this library which still uses v.2.3 and didn't have the showHint() + showHints() commands, so I've added the 'available since' level to the docs.


afshinm:
thanks @Simon-Briggs

PR #677, opened 2017-01-17
center hints and add all position combinations

Simon-Briggs:
Hi,  

The position of the hints wasn't taking in to account the size of the hint itself, so was  slightly off center.

I also wanted to center the height of the hints, so added options for that.


afshinm:
this is perfect, thanks @Simon-Briggs 

I will merge it for the next version.

PR #678, opened 2017-01-22
Condition attribute for steps and edited redundant loop

stijn1989:
I use intro.js when users push the 'help' button on the page. But the page is sometimes displayed on smartphones, tablets or computers. So some steps aren't needed to be displayed because other elements are hidden on small devices.

That's why I've build condition in steps. You can do it in JSON or attribute "data-condition". I've included an example for each of them.

Next I've read the TODO in the method "_introForElement". So I've decided to give it a try. The loop is easy. It loops through all the data-intro elements. The ones with data-step are kept in place but the ones without data-step are placed when there is space or replaced if an other data-step needs to be there.
There is an array that keep track which steps are occupied and which are occupied to but can't switch place.

### DIFF
--- a/example/condition/html-attributes.html
+++ b/example/condition/html-attributes.html
@@ -1,0 +1,74 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Basic usage</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="description" content="Intro.js - Better introductions for websites and features with a step-by-step guide for your projects.">
+    <meta name="author" content="Afshin Mehrabani (@afshinmeh) in usabli.ca group">
+
+    <!-- styles -->
+    <link href="../assets/css/bootstrap.min.css" rel="stylesheet">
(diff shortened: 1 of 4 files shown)


afshinm:
This is very interesting. I was thinking about adding the same concept using https://github.com/usablica/parserjs. My only concern is that `return new Function()` might have some security issue because this can evaluate any function, e.g. alert()

stijn1989:
Yes I've read that eval() is very risky with security and then I found this "new Function" workaround. But if the condition doesn't come from 3rd party sources, why should you worry about it? The JSON or attributes are static programmed.

afshinm:
What do you think @bozdoz

bozdoz:
Looks better suited for the programmatic method, using onbeforechange, or something like that.  I think JavaScript should be in JavaScript, not HTML, generally.  I don't really understand why you'd want to add it within HTML; but maybe I'm not the right person to evaluate this.  I suppose it could help some editor of a WordPress blog, or some other content-editor, where the user only has access to HTML.  That's the best use-case I can think of.  I am also unsure of the risk involved; I agree that if they're static, why worry.  Though I have heard, of course that eval is evil.

PR #679, opened 2017-01-23
api to programatically show hint dialog

Simon-Briggs:
Hi, 

I wanted to be able to programatically open up the hint dialog, so slightly modified the _hintClick method to make it easier to call from elsewhere.

### DIFF
--- a/docs/docs/hints/api.md
+++ b/docs/docs/hints/api.md
@@ -152,6 +152,24 @@ introJs().hideHint(0); //you should pass the id of the hint (the `data-step` att
 
 -----
 
+-----
+
+##### introJs.showHintDialog(stepId)
+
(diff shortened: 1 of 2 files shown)


stijn1989:
Aren't hints supposed to be clicked? If you need to show a step, can't you use the start() method?

Simon-Briggs:
I've been using the start method for a brief overview of the page, then hints afterwards to help users along. But I found that just a hint icon wasn't obvious enough, especially for the first one, so wanted to show them automatically.

afshinm:
This is a good idea, I like it. I'm going to test it locally and then merge it for the next version. 

Thanks!

afshinm:
merged.

PR #685, opened 2017-02-24
Try to fix methods being called on `null`

### LINKED ISSUE #638: prevTooltipButton/nextTooltipButton is sometimes null
Encountering errors like

```
Unable to get property 'removeAttribute' of undefined or null reference
```

at https://github.com/usablica/intro.js/blob/v2.3.0/intro.js#L1012

or 

```
Unable to get property 'tabIndex' of undefined or null reference
```

at https://github.com/usablica/intro.js/blob/v2.3.0/intro.js#L856

Is it missing some null checking?

PikachuEXE:
Fixes #638 
Has been using https://github.com/PikachuEXE/intro.js/tree/fix/null-reference for months (which is based in 2.3.0)

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -868,10 +868,10 @@
         if (oldHelperNumberLayer) oldHelperNumberLayer.style.opacity = 1;
 
         //reset button focus
-        if (nextTooltipButton.tabIndex === -1) {
+        if (typeof nextTooltipButton !== "undefined" && nextTooltipButton != null && nextTooltipButton.tabIndex === -1) {
           //tabindex of -1 means we are at the end of the tour - focus on skip / done
           skipTooltipButton.focus();
-        } else {
+        } else if (typeof nextTooltipButton !== "undefined" && nextTooltipButton != null) {
           //still in the tour, focus on next
           nextTooltipButton.focus();
         }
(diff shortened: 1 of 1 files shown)


afshinm:
thanks for this fix. I will test it locally and then merge

Adam13531:
@afshinm hey, is there an ETA on incorporating this fix? I tested it out in my own codebase and it enabled creating tutorials with a single step (which previously wasn't working).

P.S. great work both of you. :)

afshinm:
thank you both @PikachuEXE @Adam13531 

I will review this locally and apply your patch @Adam13531!

afshinm:
done!

Adam13531:
@afshinm hey, thanks for merging in the request! The only thing I wanted to point out is that the added conditions are redundant:

```js
typeof nextTooltipButton !== "undefined" && nextTooltipButton != null
```

could just be 

```js
nextTooltipButton != null
```

PikachuEXE:
I think I meant to use 
```js
typeof nextTooltipButton !== "undefined" && nextTooltipButton !== null`
```
But
```js
nextTooltipButton != null
```
also works too.

PR #687, opened 2017-03-02
feature(refresh): reposition tooltip on screen resize

Simon-Briggs:
Hi,

I've added 1 extra call to the _placeTooltip function when the user resizes the screen, as if the DOM changes dimension,  the tooltip will not be aligned

eg: on this demo step 2 and 3, if you resize the screen till it switches between mobile mode and desktop mode, the tooltip will not be re-aligned.
http://introjs.com/example/hello-world/index.html

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -247,8 +247,7 @@
       };
 
       self._onResize = function(e) {
-        _setHelperLayerPosition.call(self, document.querySelector('.introjs-helperLayer'));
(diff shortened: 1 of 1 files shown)


millerscout:
good idea :)

afshinm:
actually this is a very good idea, thanks for this. Do you have any jsfiddle for this?

PR #700, opened 2017-04-10
feature: add index to step callbacks

xiamike:


afshinm:
I'm happy with this change but could you please update the doc as well?

PR #704, opened 2017-04-26
Easier "Next" on mobile

kramos:
Adding the option so that clicking the main text is the same as clicking the next (or done) button.  Defaults to false to preserve normal behaviour by default.  This makes things work better on mobile when the buttons are too small.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -77,7 +77,9 @@
       /* Hint button label */
       hintButtonLabel: 'Got it',
       /* Adding animation to hints? */
-      hintAnimation: true
+      hintAnimation: true,
+      /* Click the text == next/done button */
+      textClickSkip: false
     };
   }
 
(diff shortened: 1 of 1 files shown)


afshinm:
Thanks for this, do you have any screenshots / online example of this change?

kramos:
Sure: 
* Go here https://safetychecker.herokuapp.com/
* Click help
* Click on the main text in the tip, doing so will do the same as the Next button until the last tip when it does the same as the Done button.

As you will see in the page source, textClickSkip = true:
`onclick="javascript:introJs().setOptions({ showStepNumbers: false, textClickSkip: true}).start();"`

afshinm:
thanks for this @kramos but I think we need a better way to enable mobile-users to use Introjs. Actually, I'm currently working on a feature that makes the buttons easier to use on mobile / tablet devices. Thanks again!

PR #705, opened 2017-05-06
Update hint's options.md documentation.

wdjunaidi:
This is to include `hints` options for creating hint programmatically.

Note: I couldn't find how the documentation website is generated out of the md files. So I can't tell whether it will looks good or not.


afshinm:
thanks!

PR #708, opened 2017-05-10
Redundant for loop removed and parseInt Replaced with unary +

vivek12345:
@afshinm 
- There was a redundant for loop which was checking all the items which do not have a data-step in them and then was pushing them at the end of intro items array.

- My PR replaces that with one loop which pushes the items with data-step into the array at the specified data-step position and all other data-intro items without data-step at the end of the array.
- I have also replaced the call to parseInt with a unary + operator.

Please review and merge.


vivek12345:
@afshinm Is there a problem with this change.Can you please review and merge this if everything is correct.

SirenHound:
I am curious what would happen if the code collected say, 3 elements, and some missing 'steps':
```
<div data-step="1"  data-intro="this is the first step"></div>
<div  data-intro="whatever"></div>
<div data-step="3"  data-intro="I think this is the third step"></div>
```
If I am reading the code correctly as it stands the order will put "whatever" into step 2, whereas this pull request would append it to step 4.

I'm not sure which is the expected behaviour, but it would seem to me that the first way is better.

vivek12345:
@SirenHound Both the codes before and after my change add the elements without data-step at the end which as per your example would append it to step 4.
My change removes the redundant for loop.

SirenHound:
No I don't think so, the data-step values are added to the array explicitly by fetching the value on line 159:
``` var step = parseInt(currentElement.getAttribute('data-step'), 10);```
and setting it into the array on line 168:
```         introItems[step - 1] = { ...```
which could leave gaps in the introItems array.
The loop at line 189:
``` 
while (true) {		
       if (typeof introItems[nextStep] == 'undefined') {		
           break;		
      } else {		
           nextStep++;		
      }		
}
```
finds the first gap, which is later used (line 203) to assign the step-less intro.

I don't have time to set up an example right now, but I hope I explained it well enough.

vivek12345:
@SirenHound Check this jsfiddle, it is using Intro.js and behaves the way I mentioned in my previous comment.

afshinm:
hey guys, apologies for the delay. thanks @vivek12345 and @SirenHound.

This change looks awesome but it has the problem @SirenHound mentioned. Any plans to fix that?

PR #714, opened 2017-05-28
Update introjs-rtl.css

mdf092:
The progress bullets in RTL version were started form left and now start from right for RTL languages


afshinm:
thanks!

PR #721, opened 2017-06-30
fix typo in hints api doc

cncolder:
`introJs.hideHint`


afshinm:
well, it is not a typo, correct property is `-step`

Issue #602, opened 2016-06-29
Make IntroJS extensible


VitorVRS:
I suggest to implement some kind o plugin system for IntroJS, since there are a lot of pull requests sending new "options" adding new features, which doesn't seem necessary.

In my case, i had to fork project and add new options (just to solve my problem). I would enjoy to turn my changes into an plugin to avoid messing with IntroJS core.

Tks.

afshinm:
That's such a good idea @VitorVRS, thanks for this.

Can you please share your thoughts on this? what kind of plugin do you need or let's say what you would like to do with a plugin?

VitorVRS:
Actually, i don't know how you could do that. But just reading some PR's i could see that people create a lot of "options" which, maybe, can be useful or not.

In my case, i modified IntroJS to work with iframes ( a very specifi way) and change do work dynamic elements for step ( i knew all of the steps, but the elements were not in DOM yet); And a lot of things.

I think you should review all PR's and analyse what is better to do. Maybe there is nothing.

Thanks for your project. :100:

julienchepa:
Hi @VitorVRS,

may be you could explain more how did you do that ? 
I need to do something similar with iFrame as you explain. 

thanks in advance.

VitorVRS:
@julienchepa in my case i used before and after callback to wait for the next element appear (iframe loading), unfortunately i did not created an fork :/ 
But my whole as based on this PR: https://github.com/usablica/intro.js/pull/596

Another thing what i did: i change most of the calculations for postioning to work with iframes.

I hope i was clear enough for you.

bozdoz:
Might use more event-based logic in the library, and perhaps create some generic, class-like objects that can be extended.  

I'm thinking of how extensible the LeafletJS library is (to which I've contributed a few plugins).  

Basically, they have generic Layers (used to be Classes) that can be extended and have methods to listen to or trigger events.  Because it's so generic, any Layer can be added to a Map, and when initialized, added, or any other event happens a plugin can execute custom code.   The plugin code has access to helper functions as well, for element creation, classname manipulation, positioning, etc.

I would be interested to hear a few more cases in which plugins would make sense, and we could develop an extensible system with those cases in mind.

PR #725, opened 2017-07-21
Added buttonClass to options

cdownie:
buttonClass value is initially set to the default value 'introjs-button'. Consumers now have the option to set their own class for all buttons used during the tour.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -77,7 +77,9 @@
       /* Hint button label */
       hintButtonLabel: 'Got it',
       /* Adding animation to hints? */
-      hintAnimation: true
+      hintAnimation: true,
+      /* additional classes to put on the buttons */
+      buttonClass: "introjs-button"
     };
   }
 
@@ -1015,7 +1017,7 @@
 
       //skip button
       var skipTooltipButton = document.createElement('a');
-      skipTooltipButton.className = 'introjs-button introjs-skipbutton';
+      skipTooltipButton.className = this._options.buttonClass + ' introjs-skipbutton ';
       _setAnchorAsButton(skipTooltipButton);
       skipTooltipButton.innerHTML = this._options.skipLabel;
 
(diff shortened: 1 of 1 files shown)


afshinm:
hey @cdownie, thanks!

I think you can use `tooltipClass` to add the class name and select the buttons using CSS selectors. Let me know what you think.

cdownie:
I think you could do that as well. 

In our case, it seemed simpler to add our existing button classes vs adding/editing new styles specifically for introjs. We essentially were able to add 'stock bootstrap' classes on the component and everything fell in line with our theme at that point.

bozdoz:
I like the idea of adding class names, but I don't like the idea of changing class names (because the introjs css would break); I just can't see it being used by many people.

But this, maybe:

```javascript
skipTooltipButton.className = 'introjs-button introjs-skipbutton';
if (this._options.btn_class) {
  _addClass(skipTooltipButton, this._options.btn_class);
}
```

Still awkward.  Maybe there's a better way to offer this level of customization without options.  For example, if we're thinking about making introjs extensible, #602, we could have a method for creating a button:

```javascript
IntroJS = {
  Button : function (html) {
    var btn = document.createElement('a');
    btn.setAttribute('role', 'button');
    btn.tabIndex = 0;
    btn.innerHTML = html;
    // use IntroJS utility functions to manipulate element
    IntroJS.utils.addClass(btn, 'introjs-button');
    this.container = btn;
  }
}
```

And then developers could overwrite the button creation method like this (this is an **extremely** simplified example):

```javascript
// save old function
IntroJS.originalButton = IntroJS.Button;
// overwrite
IntroJS.Button = function (html) {
  // call parent function
  IntroJS.originalButton.call(this, html);
  IntroJS.utils.addClass(btn, 'btn-primary btn-lg');
};
```

Ideally we'd have a whole system where developers can easily and actually extend functions (rather than overwrite them).  It might be helpful, but it also might be excessive.  Just an idea. :D

cdownie:
The intention in this case was indeed to overwrite, replace, and 'break' the existing styles. Without removing the existing introjs-button style, further work would have needed to be done to remove and reset the existing styles added by the stock class.

With this method I am able to apply multiple classes to a button (ie 'btn btn-link btn-sm') and achieve the desired look without modifying existing styles. If a user wished, they could include introjs-button in their list of buttonOptions and keep the existing styles

styledbuttons

bozdoz:
Interesting. Thanks for the use case and example!  I can see how this would be useful, but, again, it would be great to be able to get to a position where the entire library is more flexible, more modular, yet still widely use-able with default settings.  This will be a good example to keep in mind.

PR #727, opened 2017-07-25
Makes introjs work with transformed elements

binford2k:
Prior to this change, any time parent elements introduced a stacking
context with certain positioning or with transforms, introjs would just
forcibly remove those. For something like opacity, that may not be a big
deal, but when transforms are an integral part of page styling, you
can't just disable them and expect things to render properly.

In those cases, now we clone the highlighted element and plop it right
on top of the highlighting layer--sidestepping the whole z-index problem
to begin with.

This also yanks out the alternate method for positioning, since that
also did not take transforms into account. Now the highlight will be
positioned properly on the page, even when parents are scaled or
translated.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -1131,7 +1131,8 @@
       nextTooltipButton.focus();
     }
 
-    _setShowElement(targetElement);
+    //Pass in whichever helperlayer we're using because _removeShowElement strips the child node classes
+    _setShowElement(targetElement, helperLayer || oldHelperLayer);
 
     if (!_elementInViewport(targetElement.element) && this._options.scrollToElement === true) {
       var rect = targetElement.element.getBoundingClientRect(),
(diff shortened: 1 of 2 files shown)


afshinm:
Very nice @binford2k. 

Could you provide an example / usage for this change? I noticed you are using it for a slider?

binford2k:
Sure. Create an element and transform it. For example, skew, rotate, scale, etc. Attach an introjs tour to it, and you'll see that the transforms disappear while the popup is live. This corrects for that.

See a demo at https://codepen.io/anon/pen/yPgLbJ

bozdoz:
The demo does not appear to do what I assume this PR is supposed to do.  Am I missing something?  It looks like when I click on the link in the CodePen example the entire page loses its transformations.  Is that correct?

binford2k:
That's what this PR *fixes*.

The codepen is stock intro.js

maxkuzmin:
@bozdoz @binford2k hi guys any update for this? I have issue with transform too, so this pr looks useful for me.

PR #736, opened 2017-08-15
cleaner removing of empties in introItems array

SirenHound:
Feedback welcome.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -215,13 +215,8 @@
     }
 
     //removing undefined/null elements
-    var tempIntroItems = [];
-    for (var z = 0; z < introItems.length; z++) {
-      introItems[z] && tempIntroItems.push(introItems[z]);  // copy non-empty values to the end of the array
-    }
-
-    introItems = tempIntroItems;
-
+    introItems = introItems.filter(function(v){return v;});
+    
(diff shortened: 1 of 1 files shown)


vivek12345:
This looks good to me.I like it.Just one question, will this filter function work on all browsers being supported by intro at the moment?

SirenHound:
Good point, I keep forgetting about browser compatibility. I don't know what versions of browsers IntroJs is supporting but it should work on all major current browsers.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

bozdoz:
Would be a good idea for the project to explicitly list which browsers and versions the project supports.  Maybe even write a test to ensure it.

@afshinm

bozdoz:
Also, this PR should work for IE9+

afshinm:
That's a great idea @bozdoz. A test would be perfect to show the browser compatibility.

I'd prefer not to add something like `Array.filter()` for now. Let's do that test and then merge this.

Looks like `filter` works for IE9+: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

PR #739, opened 2017-08-18
Adding an option to prevent displaying the next / previous step

afshinm:


afshinm:
I have to merge the other PR because I need that version immediately. @bozdoz let's review this and add that for 2.7.1.

afshinm:
Thanks @bozdoz

PR #741, opened 2017-08-22
Add ability to force a tooltip's position

Josh5:
Small windows (such as mobile) are having position being forced to floating due to auto positioning.
There are cases where this is wrong and we would like it to be forced to attach to the top or bottom of the element.
This commit allow us to simply add ".forced" to the end of the position string in order to completely bypass the auto positioning of IntroJs.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -567,7 +567,9 @@
 
     currentTooltipPosition = this._introItems[this._currentStep].position;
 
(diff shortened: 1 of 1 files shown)


afshinm:
hey @Josh5. thanks for this. Looks like a good idea but to be honest, I don't think this implementation can solve all the problems.

I'm currently working on fixing the mobile issues. Feel free to add more @bozdoz .

PR #745, opened 2017-08-25
Support line feed in data-intro

zhouyuexie:
Our code:

```html
<div data-intro="line1\nline2\n" data-step="1">...</div>
```

Previous code display:

```html
<div class="introjs-tooltiptext">line1\nline2\n</div>
```

**Now:**

```html
<div class="introjs-tooltiptext">
  <p>line1</p>
  <p>line2</p>
</div>
```

Although can wirte code like:

```html
<div data-intro="<p>line1</p><p>line2</p>" data-step="1">...</div>
```

But it's too anti human.

Thanks.


afshinm:
Thanks for this but if you use the programmatic version (like passing JSON) you would be able to include HTML tags.

zhouyuexie:
I see...

Thanks for you replied.

afshinm:
Thank _you_ for helping the project!

Issue #170, opened 2013-07-30
Keyboard arrow keys do not work as navigation in Opera


jamesbloomer:
Ubuntu 12.04
Opera 12.16

The buttons work fine but the keyboard bindings to the arrow keys do not work, nothing happens when they are pressed.

PR #753, opened 2017-09-04
Changed keyCode to check for modern `code`, then `which` then `charCode`

bozdoz:
Should resolve any key-related issues, such as #170


afshinm:
I noticed I cannot move between steps after merging this. Could you have a look @bozdoz ?

afshinm:
`code` returns something like ArrowLeft, ArrowRIght, etc.

PR #754, opened 2017-09-05
changes to fix overlay issue in case of multiple intro flows. 

mkumary:
In case of multiple intro on same page overlay only covers the element which makes no sense and UI looks odd.
made changes to fix this issue and always overlay body.


afshinm:
Hey @mkumary. thanks for this, but we use `targetElm` almost everywhere and this simple change might cause some side effects.

Could you provide an example of the problem you've mentioned?

PR #756, opened 2017-09-11
Rebase our fork against upstream master

joelmichael:
A lot has changed in this library so I had to rebase it. I think this preserves all of our changes fine.

However we need to seriously consider moving off of our custom changes and just using the official library. Alternatively make a PR for our changes to the official library. Maintaining this library in this way is not sustainable.


joelmichael:
Sorry I meant to make this PR for our repo.

PR #763, opened 2017-09-26
Re-added PR #80 for groups

bozdoz:
I omitted changes to css and minified files, and added an example.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -159,10 +159,9 @@
 
     } else {
       //use steps from data-* annotations
-      var allIntroSteps = targetElm.querySelectorAll('*[data-intro]');
       var elmsLength = allIntroSteps.length;
       var disableInteraction;
-
+      
       //if there's no element to intro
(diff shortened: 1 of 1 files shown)


afshinm:
Hmm, nice. The thing is, we can already pass a parent element and start the intro for a specific container rather than document.body.

Do you think these two have conflict with each other?

bozdoz:
I think it would find the group matches within the selected element, wouldn't it? Don't think it would conflict.

afshinm:
Do you think this PR can address an issue like this: https://stackoverflow.com/questions/46679734/intro-js-how-to-select-multiple-element-navigate-between-pages-with-continuous ?

bozdoz:
That's quite an involved StackOverflow question, so I am not sure.  My best guess use-case is to run one introjs after another, like in the example I posted with it: one iterated even numbered sections and the other iterated the odd, and there was a button for each.

PR #764, opened 2017-09-26
Basic jsHint implementation

bozdoz:
I think this would go a long way to help with future Pull Requests.  We could add this repo to https://travis-ci.org and it will auto check for any basic JavaScript issues.

Future testing steps could be introducing [Jest](https://facebook.github.io/jest/) for basic unit testing, and CasperJS for simulated implementation tests.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -113,13 +113,12 @@
    */
   function _introForElement(targetElm) {
     var introItems = [],
-        self = this,
-        i = 0;
+        self = this;
 
     if (this._options.steps) {
       //use steps passed programmatically
-      for (var stepsLength = this._options.steps.length; i < stepsLength; i++) {
-        var currentItem = _cloneObject(this._options.steps[i]);
+      _forEach(this._options.steps, function (step) {
+        var currentItem = _cloneObject(step);
 
         //set the step
         currentItem.step = introItems.length + 1;
@@ -153,14 +152,13 @@
 
         if (currentItem.element !== null) {
           introItems.push(currentItem);
-        }
-      }
+        }        
+      }.bind(this));
 
     } else {
       //use steps from data-* annotations
       var allIntroSteps = targetElm.querySelectorAll('*[data-intro]');
       var elmsLength = allIntroSteps.length;
-      var currentElement;
       var disableInteraction;
 
       //if there's no element to intro
@@ -168,13 +166,10 @@
         return false;
       }
 
-      //first add intro items with data-step
-      for (i = 0; i < elmsLength; i++) {
-        currentElement = allIntroSteps[i];
-
+      _forEach(allIntroSteps, function (currentElement) {
         // skip hidden elements
         if (currentElement.style.display === 'none') {
-          continue;
+          return;
         }
 
         var step = parseInt(currentElement.getAttribute('data-step'), 10);
@@ -197,14 +192,13 @@
             disableInteraction: disableInteraction
           };
         }
-      }
+      }.bind(this));
 
       //next add intro items without data-step
       //todo: we need a cleanup here, two loops are redundant
       var nextStep = 0;
-      for (i = 0; i < elmsLength; i++) {
-        currentElement = allIntroSteps[i];
 
+      _forEach(allIntroSteps, function (currentElement) {
         if (currentElement.getAttribute('data-step') === null) {
 
           while (true) {
@@ -213,7 +207,7 @@
             } else {
               nextStep++;
             }
-          }
+          } 
 
 
           if (typeof (currentElement.getAttribute('data-disable-interaction')) !== 'undefined') {
@@ -233,7 +227,7 @@
             disableInteraction: disableInteraction
           };
         }
-      }
+      }.bind(this));
     }
 
     //removing undefined/null elements
@@ -397,13 +391,12 @@
     this._direction = 'forward';
 
     if (typeof (this._currentStepNumber) !== 'undefined') {
-        for( var i = 0, len = this._introItems.length; i < len; i++ ) {
-            var item = this._introItems[i];
-            if( item.step === this._currentStepNumber ) {
-                this._currentStep = i - 1;
-                this._currentStepNumber = undefined;
-            }
+      _forEach(this._introItems, function (item, i) {
+        if( item.step === this._currentStepNumber ) {
+          this._currentStep = i - 1;
+          this._currentStepNumber = undefined;
         }
+      }.bind(this));
     }
 
     if (typeof (this._currentStep) === 'undefined') {
@@ -516,19 +509,16 @@
 
     //remove overlay layers from the page
     var overlayLayers = targetElement.querySelectorAll('.introjs-overlay');
-    var i;
 
-    if (overlayLayers && overlayLayers.length > 0) {
-      for (i = overlayLayers.length - 1; i >= 0; i--) {
-        //for fade-out animation
-        var overlayLayer = overlayLayers[i];
+    if (overlayLayers && overlayLayers.length) {
+      _forEach(overlayLayers, function (overlayLayer) {
         overlayLayer.style.opacity = 0;
         window.setTimeout(function () {
           if (this.parentNode) {
             this.parentNode.removeChild(this);
           }
         }.bind(overlayLayer), 500);
-      }
+      }.bind(this));
     }
 
     //remove all helper layers
@@ -558,11 +548,9 @@
 
     //remove `introjs-fixParent` class from the elements
     var fixParents = document.querySelectorAll('.introjs-fixParent');
-    if (fixParents && fixParents.length > 0) {
-      for (i = fixParents.length - 1; i >= 0; i--) {
-        _removeClass(fixParents[i], /introjs-fixParent/g);
-      }
-    }
+    _forEach(fixParents, function (parent) {
+      _removeClass(parent, /introjs-fixParent/g);
+    });
 
     //clean listeners
     if (window.removeEventListener) {
@@ -925,7 +913,6 @@
         oldHelperLayer = document.querySelector('.introjs-helperLayer'),
         oldReferenceLayer = document.querySelector('.introjs-tooltipReferenceLayer'),
         highlightClass = 'introjs-helperLayer',
-        i,
         nextTooltipButton,
         prevTooltipButton,
         skipTooltipButton;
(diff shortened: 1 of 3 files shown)


intro.js
```diff
@@ -484,8 +495,8 @@
     // calling onbeforeexit callback
     //
     // If this callback return `false`, it would halt the process
-    if (this._introBeforeExitCallback != undefined) {
-      continueExit = this._introBeforeExitCallback.call(self);
```
bozdoz: `self` was undefined

package.json
```diff
@@ -8,7 +8,12 @@
     "type": "git",
     "url": "https://github.com/usablica/intro.js"
   },
+  "scripts": {
+    "test": "npm run test:jshint",
```
bozdoz: run with `npm test`

intro.js
afshinm: we should pass `this`

intro.js
```diff
@@ -136,34 +137,40 @@
...
+      if (elmsLength < 1) {
         return false;
       }
 
+      i = 0;
+
       //first add intro items with data-step
-      for (var i = 0, elmsLength = allIntroSteps.length; i < elmsLength; i++) {
-        var currentElement = allIntroSteps[i];
+      for (; i < elmsLength; i++) {
```
afshinm: I think it makes more sense to define `i` inside the for (statement 1)

intro.js
```diff
@@ -136,34 +137,40 @@
...
 
         var step = parseInt(currentElement.getAttribute('data-step'), 10);
 
-        var disableInteraction = this._options.disableInteraction;
-
-        if (typeof (currentElement.getAttribute('data-disable-interaction')) != 'undefined') {
+        if (typeof (currentElement.getAttribute('data-disable-interaction')) !== 'undefined') {
           disableInteraction = !!currentElement.getAttribute('data-disable-interaction');
+        } else {
+          disableInteraction = this._options.disableInteraction;
```
afshinm: this is the base case and can be defined outside the if. like removing `var disableInteraction;` and initiating with  `this._options.disableInteraction`

intro.js
```diff
@@ -1079,8 +1099,8 @@
       tooltipLayer.appendChild(progressLayer);
 
       //add helper layer number
-      if (this._options.showStepNumbers == true) {
-        var helperNumberLayer = document.createElement('span');
+      var helperNumberLayer = document.createElement('span');
+      if (this._options.showStepNumbers === true) {
```
afshinm: Why is it necessary?

intro.js
bozdoz: I agree, there might be an option in jshint to change how it interprets this.  Basically, you can't have two forloops in the same block-scope re-defining the same variables. So:

```javascript
for (var i = 0...) {} 
// and then
for (var i = 0...) {}
```

Is re-defining `i`, which jshint disagrees with, and though I think it's good practice, I also agree with you that it makes things look a little more awkward (though maybe there's just a better way to do it).  ES6 recommends using `let` instead of `var`, and we can start doing that in a later version.  

I suppose for this we could wrap it in a closure? Or we can look into changing the jshint options.  What do you think?

intro.js
bozdoz: I can't remember this one in context, but I believe it was something like this:

```javascript
if (1 === 1) {
  var a = 3;
}
// a is defined within an if-block
return a;
```

While it technically "works", it's bad style and makes it more ambiguous; generally you want to define all your variables at the top of the block-scope that you'll be using them in; at the top of functions, at the top of if statements, at the top of forloops, and don't use them outside of those block scopes.  I think it's good practice for readability if nothing else.

intro.js
bozdoz: And actually maybe the solution here is just to remove the `var`:

```javascript
for (i = 0, i < elmsLength; i++)
```

That probably solves it and makes it just as readable

intro.js
bozdoz: I did below. :D

bozdoz:
I changed those for loop variables; just removed the `var` and rebased it with master.

intro.js
```diff
@@ -113,11 +113,12 @@
    */
   function _introForElement(targetElm) {
     var introItems = [],
-        self = this;
+        self = this,
+        i = 0;
```
afshinm: Can we define this inside the loop?

intro.js
```diff
@@ -113,11 +113,12 @@
...
 
     if (this._options.steps) {
       //use steps passed programmatically
-      for (var i = 0, stepsLength = this._options.steps.length; i < stepsLength; i++) {
+      for (var stepsLength = this._options.steps.length; i < stepsLength; i++) {
```
afshinm: Here I meant

intro.js
```diff
@@ -150,34 +151,38 @@
...
       //if there's no element to intro
-      if (allIntroSteps.length < 1) {
+      if (elmsLength < 1) {
         return false;
       }
 
       //first add intro items with data-step
-      for (var i = 0, elmsLength = allIntroSteps.length; i < elmsLength; i++) {
-        var currentElement = allIntroSteps[i];
+      for (i = 0; i < elmsLength; i++) {
```
afshinm: And I think `i` should only be visible to this for-loop block. If we remove the previous `var i` and define `i` here.

intro.js
```diff
@@ -197,23 +202,24 @@
       //next add intro items without data-step
       //todo: we need a cleanup here, two loops are redundant
       var nextStep = 0;
-      for (var i = 0, elmsLength = allIntroSteps.length; i < elmsLength; i++) {
-        var currentElement = allIntroSteps[i];
+      for (i = 0; i < elmsLength; i++) {
```
afshinm: and here as well, `var i`.

intro.js
```diff
@@ -508,18 +516,19 @@
 
     //remove overlay layers from the page
     var overlayLayers = targetElement.querySelectorAll('.introjs-overlay');
+    var i;
```
afshinm: same as previous comment, defining `var i` inside the loop.

intro.js
```diff
@@ -919,7 +925,10 @@
         oldHelperLayer = document.querySelector('.introjs-helperLayer'),
         oldReferenceLayer = document.querySelector('.introjs-tooltipReferenceLayer'),
         highlightClass = 'introjs-helperLayer',
-        elementPosition = _getOffset(targetElement.element);
+        i,
```
afshinm: this one, `var i`.

intro.js
```diff
@@ -1045,15 +1056,19 @@
       var ulContainer = document.createElement('ul');
       ulContainer.setAttribute('role', 'tablist');
 
-      for (var i = 0, stepsLength = this._introItems.length; i < stepsLength; i++) {
+      var anchorClick = function () {
+          self.goToStep(this.getAttribute('data-stepnumber'));
+      };
+
+      i = 0;
```
afshinm: this one as well, `var i`.

intro.js
```diff
@@ -1569,11 +1586,13 @@
    * @method _startHint
    */
   function _populateHints(targetElm) {
-    var self = this;
+    var i = 0,
```
afshinm: this `var i`, too.

intro.js
bozdoz: I think you've come across the issue exactly: `var` would not be only visible to that for-loop block; that's why jshint complains that `i` is being redefined.  We could move to ES6 and do `let`, which might not be a great idea yet, or, we could wrap these for loops in a closure, maybe even have a helper function for iterating arrays; that way `var` would be only visible in that block.  Let me know!

afshinm:
@bozdoz could you please address the issues I have mentioned? If you don't have enough time, I can work on this.

intro.js
bozdoz: @afshinm Here's what I was waiting on.  Here are some suggested ways to address this issue:

1. wrap the loops in closures:
```javascript
(function () {
  for (var i = 0; len = steps.length; i < len; i++) { 
    // etc
  }
})();
```
2. helper function to deal with loops:
```javascript
_forEach(steps, forEachFnc, endLoopFnc);
```

And I don't yet recommend introducing the ES6 variant `let`, but at some point in the future it might be a good idea.

bozdoz:
I wrote a comment on one of your comments above, @afshinm. Sorry if you missed it there.

bozdoz:
There might be some good resources on this actually.  Found this one; somewhat related: https://stackoverflow.com/questions/19412727/why-is-jshint-saying-variable-already-defined-in-this-if-statement

bozdoz:
Here's another that's more specific to the for loops: https://stackoverflow.com/questions/25102262/is-there-a-way-to-silence-jshints-variable-is-already-defined-warning

In there, one solution is to use different letter variables.  So we could do `var i`, then `var j`, `var k`, whatever.  But it is important to note that because of variable hoisting, what's actually happening is that both `i` variables are defined as soon as the function is executed:

```javascript
function () {
  var i = 0;
  var i = 0; // kind of silly
  // for loop here
```

intro.js
afshinm: @bozdoz ah I see. Let's do that _forEach thing, makes more sense to me. +1 for `let` but in the future.

afshinm:
Very nice @bozdoz, could you rebase it? I think we can merge it.

afshinm:
@bozdoz awesome! I'm a bit worried about other PRs though. Let's merge this one and manually rebase others.

intro.js
```diff
@@ -251,9 +254,6 @@
       //then, start the show
       _nextStep.call(self);
 
-      var skipButton     = targetElm.querySelector('.introjs-skipbutton'),
-          nextStepButton = targetElm.querySelector('.introjs-nextbutton');
```
afshinm: @bozdoz I assume we were not using this variables, right?

intro.js
```diff
@@ -1370,6 +1371,27 @@
...
+  * @param {Array} arr
+  * @param {Function} forEachFnc
+  * @param {Function} completeFnc
+  * @return {Null}
+  */
+  function _forEach(arr, forEachFnc, completeFnc) {
+    // in case arr is an empty query selector node list
+    if (arr) {
+      for (var i = 0, len = arr.length; i < len; i++) {
+        forEachFnc(arr[i], i);
```
afshinm: what about `forEachFnc.call(this, arr[i], i);`? just to pass Introjs instance to the fn. Same for completeFnc.

intro.js
bozdoz: Correct, jshint said they were unused.

intro.js
bozdoz: Some of them don't need it, and because it's a private function, this might be ambiguous and we control all the calls, so we can use `.bind(this)` where appropriate.  Right now most of them look like this:

```javascript
_forEach(arr, function (item) {
  // do something here within the same context
}.bind(this));
```

bozdoz:
Hopefully this will help with PR's.  I do realize most of them would have to rebase, but hopefully the good ones are concise enough not to need much work to rebase.  After this is merged, we can sign up for [Travis CI](https://travis-ci.org/) and it will automatically check PR's for jshint recommended styles and obvious errors.  After that we can go further and maybe use [jest](https://facebook.github.io/jest/) and build some more practical tests: make sure basic examples/functions work properly.  At that point we'll know when a PR breaks something.

intro.js
afshinm: ah I see.

afshinm:
Let's go then

PR #768, opened 2017-09-28
Fixes Autoposition in some cases

### LINKED ISSUE #681: Bug: Calculating auto position removes top/bottom placement options incorrectly
Easy to reproduce, just open the auto position demo page and shrink your window enough.

image

Caused by this line in `_determineAutoPosition` function.
```javascript
661:  if (targetOffset.left + tooltipWidth > windowSize.width || ((targetOffset.left + (targetOffset.width / 2)) - tooltipWidth) < 0) {
```
What is the purpose of the second part of the condition? I don't see any reason to have it there. Even if the tooltip had been positioned in something like `bottom-center` you would have to divide `tooltipWidth/2`.

I think this should be removed, it basically breaks the `auto` positioning on mobile devices.

bozdoz:
Removed the latter half of an if statement in _determineAutoposition, as recommended in #681.  

Added 'introjs-' position to the tooltip className so users can customize css by that if they wish.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -628,7 +628,7 @@
           targetLeft = targetOffset.left;
 
       if (targetLeft > screenThird) {
-        // hint is less than 1/3 of the screen width
+        // hint is greater than 1/3 of the screen width
         if (targetLeft < (screenWidth - screenThird)) {
           // hint is less than 2/3 of the screen width
           currentTooltipPosition = 'bottom-middle-aligned';
--- a/intro.js
+++ b/intro.js
@@ -621,6 +621,8 @@
     tooltipOffset = _getOffset(tooltipLayer);
     windowSize    = _getWinSize();
 
+    _setClass(tooltipLayer, 'introjs-' + currentTooltipPosition);
+
     switch (currentTooltipPosition) {
       case 'top':
         arrowLayer.className = 'introjs-arrow bottom';
@@ -770,7 +772,7 @@
 
     // Check if the width of the tooltip + the starting point would spill off the right side of the screen
     // If no, neither bottom or top are valid
-    if (targetOffset.left + tooltipWidth > windowSize.width || ((targetOffset.left + (targetOffset.width / 2)) - tooltipWidth) < 0) {
+    if (targetOffset.left + tooltipWidth > windowSize.width) {
       _removeEntry(possiblePositions, "bottom");
       _removeEntry(possiblePositions, "top");
     } else {
--- a/intro.js
+++ b/intro.js
@@ -609,36 +609,14 @@
 
     currentTooltipPosition = this._introItems[this._currentStep].position;
 
-    if (currentTooltipPosition != "floating") { // Floating is always valid, no point in calculating
-      if (currentTooltipPosition === "auto") {
-        currentTooltipPosition = _determineAutoPosition.call(this, targetElement, tooltipLayer);
-      } else {
-        currentTooltipPosition = _determineAutoPosition.call(this, targetElement, tooltipLayer, currentTooltipPosition);
-      }
+    // Floating is always valid, no point in calculating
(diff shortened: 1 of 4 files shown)


bozdoz:
With the scroll position included in autopositioning, the tooltip would appear on screen in situations where it might not have if the user scrolled:

Normal position
top-align

Autopositioning on the bottom, after scroll:
auto-bottom-align

afshinm:
This is awesome @bozdoz! I'm going to test it locally.

afshinm:
I noticed a few issues with this branch but looks good in general. I'm testing on Firefox.

2017-10-08-141940_1016x1671_scrot
2017-10-08-141933_1016x1671_scrot

bozdoz:
I'll take a look

bozdoz:
Something looks very wrong there @afshinm.  I wondered if maybe somehow there was something off with Firefox, but I'm testing on it now and can't reproduce your screenshots.  I notice that they aren't right-aligned, first of all (they don't have the arrow on the top-right).  And they're obviously positioned off the page.  I'm trying Firefox, on responsive design mode, with Google Nexus 5 selected (though I've tried a few others too).  I'm not sure what is happening.  Any suggestions, ideas?

afshinm:
Hmm, I'm using i3 wm and I didn't test that on responsive mode, the window
width was not big enough to fit the page.

I will test it again on Google Chrome.

bozdoz:
Did you display the hints and then shrink the page?

bozdoz:
Nevermind, I reproduced the error by shrinking screen size, as opposed to using responsive mode.  I'll take another look now.

intro.js
```diff
@@ -609,37 +609,15 @@
...
 
-    if (currentTooltipPosition != "floating") { // Floating is always valid, no point in calculating
-      if (currentTooltipPosition === "auto") {
-        currentTooltipPosition = _determineAutoPosition.call(this, targetElement, tooltipLayer);
-      } else {
-        currentTooltipPosition = _determineAutoPosition.call(this, targetElement, tooltipLayer, currentTooltipPosition);
-      }
+    // Floating is always valid, no point in calculating
+    if (currentTooltipPosition != "floating") { 
+      currentTooltipPosition = _determineAutoPosition.call(this, targetElement, tooltipLayer, currentTooltipPosition);
```
bozdoz: _determineAutoPosition already checks for 'auto'

intro.js
```diff
@@ -864,10 +853,66 @@
       }
     }
 
+    // only top and bottom positions have optional alignments
+    if (['top', 'bottom'].indexOf(calculatedPosition) !== -1) {
+      calculatedPosition += _determineAutoAlignment(targetOffset.left, tooltipWidth, windowSize, desiredAlignment);
```
bozdoz: alignment is calculated after position, when we know whether the auto-position will be top or bottom

bozdoz:
Testing is already proving its worth! :D

gilesbutler:
Any idea when this will be merged and released guys? :)

afshinm:
Looks good to me, feel free to merge @bozdoz

gilesbutler:
Thanks for the update @afshinm and @bozdoz 🙂 do you have a rough eta on the next release at all?

intro.js
```diff
@@ -757,40 +782,44 @@
...
-        _removeEntry(possiblePositions, "bottom");
-      }
+    }
 
-      // Check for space above
-      if (targetOffset.top - tooltipHeight < 0) {
-        _removeEntry(possiblePositions, "top");
-      }
+    // Check for space above
+    if (targetOffset.top - tooltipHeight < 0) {
```
tianyong90: When the target-element is fixed at the bottom, just like many tabbar menus do, the expression will always result a true value, however, the space above is big enough……

PR #771, opened 2017-10-07
Desired position should work, when it's NOT in possible positions

metronomx:
Current logic is opposite: only positions that are present in positionPrecedence array are working. Other positions **never** work.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -802,7 +802,7 @@
 
     // If the requested position is in the list, replace our calculated choice with that
     if (desiredTooltipPosition && desiredTooltipPosition != "auto") {
-      if (possiblePositions.indexOf(desiredTooltipPosition) > -1) {
+      if (possiblePositions.indexOf(desiredTooltipPosition) == -1) {
         calculatedPosition = desiredTooltipPosition;
       }
     }


afshinm:
Hmm, seems like a weird mistake. Looks good to me. cc-ing @bozdoz

bozdoz:
I disagree with this.  If the position has been calculated to be off the screen, I do not think it should be shown in that position.  `calculatedPosition` should appear in `possiblePositions`, otherwise it would appear off-screen...  Correct me if I'm wrong.

metronomx:
Positions, other than present in positionPrecedence array (e.g. "bottom", "top", "right", "left"), NEVER work due to this part of code:
https://github.com/usablica/intro.js/blob/55f6bb746a9f48e43d4cc6b8740e28f1c4dfedf4/intro.js#L804-L808
Currently, this is the only place, where position could be set to 'bottom-middle-aligned' or 'bottom-right-aligned' and it fails.

My fix may be rough, but at least it allows user to set it on his own risk. This fix made the job for me.

bozdoz:
OK, to me, that's a different issue.  If you're desiring to position the tip on bottom but the element is at the bottom of the screen, it **should not** appear on bottom, because it **will not** be visible.

The `middle-` and `right-aligned` positions are a different issue, which I addressed here in #768 :

https://github.com/usablica/intro.js/blob/c195afa6823a1262a96730efa9189fbb4e74d678/intro.js#L624-L640

Basically it auto assigns middle and right based on which area of the screen the element appears: if the screen were split in thirds for each of the three positions.  I think this is a better solution, but by doing this it removes the possibility of setting `middle-` or `right-aligned` styles explicitly (which might be easier for the user anyway in most cases).  Another option might be to check for valid alignments, in addition to checking positions.  An element in the middle of the page, for example, might easily be able to have a tooltip aligned left, middle, or right.  We could do the same kind of check and either auto-assign or choose the desired alignment.  Let me know what you think.

metronomx:
I think it's wrong not to deny user to define position by himself. I learned it the hard way just lately: the customer insisted on bottom-centered position. To me it's better if I could try to set it, then see it look bad and then remove the option or do something else to make it look better. But now it's just too disappointing, when features, described as common options, are not working, and you can't do anything with this.

I'll try to come up with more flexible solution when I have time.

bozdoz:
I agree with the concept generally, but I also think sizes and positions are too difficult to manage with **static** values.  Most web design these days focus on responsive web design, and part of responsive design is **dynamic** sizes and positions.  

For example, one of the tests I did on #768 was to statically position the header tooltip on the top: When I was at the top of the page, the tooltip appeared at the top, as expected.  But when I scrolled down and prompted the tooltip, what should it do?  Should it appear off the screen, on top of the header?  That's what I set it to do.  And that **could** be an option, but **I think** it makes more sense to make the tooltip visible and force it to the bottom position; otherwise it would appear as though nothing happened, and I would think the user would be confused.  

So, in general, I just can't imagine most people would want this library to allow for forced positions off-screen.  I believe most of the issues and PR's deal with visibility.  So, I would suggest that we have a function for auto-positioning which chooses desired position where available, and auto-aligning which chooses desired alignment where available.  But maybe I'm misunderstanding the point of having `data-position` in the first place. Maybe it is supposed to be a forced position, and maybe that's how most users are using it.  Thoughts @afshinm ?  Should `data-position` be a preferred position or a forced position?

bozdoz:
Maybe what we need is for `data-position` to be a list of positions ordered by priority.

I'm thinking back to the project I used this library on, and I had preferred positions, but they didn't work on all screen dimensions.  A list of fall-back positions might help; and the way I interpreted `data-position` was a sort of preferred position which didn't take effect if it would be off-screen.  That worked for my project, but maybe not for everyone.  Hope that helps the conversation. :smiley: 


bitmoji

bozdoz:
Just put some more work into #768.  Now it will prefer alignment specified in `data-position`. This should be at least closer to what you're looking for, @metronomx.  If you chose `top-right-aligned` for example, and the element is so far left that it will spill off screen, it would change to `top-middle-aligned`.  So, it prefers a given alignment, but only if it's on-screen.

PR #775, opened 2017-10-11
Changed camelcase data attributes to lowercase

### LINKED ISSUE #519: data attribute names are invalid
I've been trying to debug an issue I was having locally, and I think it might be in part due to the fact that the suggested `data-` attribute for adjusting hint position is called [`data-hintPosition`](https://github.com/usablica/intro.js/blob/8979fc6569fb405c987e1d9ebbc94113b530d6d1/intro.js#L1241-L1242). 

According to the [HTML spec](https://html.spec.whatwg.org/multipage//dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes):

> A custom data attribute is an attribute in no namespace whose name starts with the string "data-", has at least one character after the hyphen, is XML-compatible, and contains no uppercase ASCII letters.

To migrate to a standards compliant approach we'd need to support both mixed-case and lowercase variants of the settings, but deprecate the mixed-case version. 

I'm happy to submit a patch with the required change. Just wanted to make sure this is a direction y'all are happy with before I commit the time to doing it.

bozdoz:
I believe this is already backwards compatible (tested in Chrome and Firefox, at least).  The browsers automatically convert camelcased attributes to lowercase, and both can be accessed successfully via `elem.getAttribute`.

See this [StackOverflow answer](https://stackoverflow.com/questions/22753629/jquery-get-html-5-data-attributes-with-hyphens-and-case-sensitivity) and the notes on how data attributes are "translated": hyphens become camel-case, and uppercase become lowercase.  So this **should** be enough.  If not, we can add a helper function which queries the data attribute and conditionally posts a deprecated warning in the console.

resolves #519

### DIFF
--- a/docs/docs/hints/attributes.md
+++ b/docs/docs/hints/attributes.md
@@ -8,4 +8,4 @@ permalink: /hints/attributes/
 You can use these HTML attributes to manage or alter hint elements.
 
  - `data-hint`: The tooltip text of hint
- - `data-hintPosition`: Optionally define the position of hint. Options: `top-middle`, `top-left`, `top-right`, `bottom-left`, `bottom-right`, `bottom-middle`, `middle-left`, `middle-right`, `middle-middle`. Default: `top-middle`
+ - `data-hintposition`: Optionally define the position of hint. Options: `top-middle`, `top-left`, `top-right`, `bottom-left`, `bottom-right`, `bottom-middle`, `middle-left`, `middle-right`, `middle-middle`. Default: `top-middle`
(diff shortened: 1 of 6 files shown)


afshinm:
Cool, could you update documentation as well? I believe we have to add the deprecation warning and / or mention that in the docs.

bozdoz:
Do we know that it isn't already backwards compatible?  I can take a look at changing the docs, maybe in a few days.

afshinm:
thanks @bozdoz

Issue #438, opened 2015-06-19
Why is there no top-middle-aligned?


jlubeck:
I see bottom-middle-aligned and bottom-right-aligned, shouldn't there be the same for top?

BKD-Digital:
that was one of the very few issues I have with this package but it did save me time in having to create one myself. Super easy to use. Time to get creative :) I haven't checked out any of the branches but let's see if anyone else has already solved this problem for us. I'm going to prioritize this one from my backlog.

It does leave the user with a sub par experience if it is the first time they visit your site. I'll let you know how I patch it up.

Maado:
You can use cookie Js 
Add a cookie value and check for it 
If it is exist dont call intro if it not exist call the intro

PR #778, opened 2017-10-20
Position tooltip arrow correctly

Adam13531:
This addresses issue #438. When `top` or some variation of `bottom` is specified, this will position the tooltip's arrow in the center of the target element rather than putting it all the way on the left side.

image

Note that I did this before realizing `bottom-middle-aligned` and `bottom-right-aligned` were valid properties, but the equivalents don't exist for `top`. I'm pretty sure that means this overrides what `bottom-X-aligned` does.

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -584,6 +584,7 @@
    */
   function _placeTooltip(targetElement, tooltipLayer, arrowLayer, helperNumberLayer, hintMode) {
     var tooltipCssClass = '',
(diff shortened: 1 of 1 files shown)


bozdoz:
This issue is addressed in #768, but I like the conditional styling of the arrow.

Adam13531:
@bozdoz closing this then.

Issue #785, opened 2017-10-30
Enabling CI


afshinm:
Hey guys,

I'd like to enable CI on the project. I would vote for Travis CI. Let let me know if you have any comments.

CC @bozdoz

PR #787, opened 2017-10-31
Adding Travis

afshinm:


.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: node_js
+
+node_js:
+  - "node"
+
+install:
+  - npm install
+
+script:
+  - make test
```
bozdoz: why not just `npm test`?  Not really sure why we even need to keep the makefile.  But that's purely opinion.  Think it makes sense to have the commands in one place.  Also think npm is standard.

afshinm: yeah sure, that works, too. I think we can remove `script`. I just thought `make test` is more standard :D

bozdoz: Yeah maybe you're right; guess travis doesn't require much.  This is from another repo I've worked on:

https://github.com/ursudio/leaflet-webgl-heatmap/blob/master/.travis.yml#L1-L3

It's just:

```
language: node_js
node_js:
  - node # latest node.js version
```

automatically does `npm install` and `npm test`

afshinm: @bozdoz so removing the extra parts?

afshinm:
closes #785

PR #792, opened 2017-11-17
Use getBoundingClientRect method to improve _determineAutoPosition

tianyong90:
`getBoundingClientRect()` is better I think, especially when the target is fixed in some case, as I mentioned in https://github.com/usablica/intro.js/pull/768

image

This worked fine on my pc and mobile 😄

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -1355,6 +1355,7 @@
    * @param {Object} tooltipLayer
    */
   function _scrollTo(scrollTo, targetElement, tooltipLayer) {
+    if (scrollTo === 'off') return;  
     var rect;
 
     if (!this._options.scrollToElement) return;
--- a/intro.js
+++ b/intro.js
@@ -792,17 +792,10 @@
     // Take a clone of position precedence. These will be the available
     var possiblePositions = this._options.positionPrecedence.slice();
 
-    var scrollTop = window.pageYOffset || document.documentElement.scrollTop;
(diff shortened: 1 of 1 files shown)


afshinm:
@bozdoz any comments?

bozdoz:
I like the idea.  I do think it's generally better/more accurate to use.  Haven't actually tested the PR, but I have no reason to believe that this is anything but an improvement. +1

RicardoVaranda:
Nice work @bozdoz & @tianyong90 ! Thanks for improving this library, @afshinm is there any estimation as to when this will be released, holding back on rolling intro.js intro production until these patches are applied.

afshinm:
thanks @tianyong90

EhsanKia:
I know it's a bit late now, but I think a much cleaner solution would've been to replace _getOffset entirely. getBoundingClientRect is definitely great, and all the hacky position calculation is unnecessary.

Simply doing
`function _getOffset(element) { return element.getBoundingClientRect(); }`

Would basically achieve what you were going for here, and even more, with a single line change.

bozdoz:
That works, @EhsanKia, but also needs the scroll position (must be a styling issue, or some other method of positioning that could be improved on; not sure).  See #801.

bozdoz:
Specifically: 

https://github.com/usablica/intro.js/pull/801/files#diff-42068bed56449c4ec84ddf8745e56b7cR2109-R2121

PR #793, opened 2017-11-21
To change the z-index value of an element with JavaScript,

skylow95:
To change the z-index value of an element with JavaScript, you would use the usual “camel casing minus the hyphen”, which is generally the norm for accessing CSS properties with JavaScript (e.g. myElement.style.zIndex = "100";)

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -1539,7 +1539,9 @@
   function _getPropValue (element, propName) {
     var propValue = '';
     if (element.currentStyle) { //IE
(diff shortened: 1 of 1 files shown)


bozdoz:
Not sure I understand.  Is this fixing a bug?  Or is this fixing something to make it more standard practice?  Is there any risk of merging this?  I'm curious, because I haven't noticed an issue around this piece of code before.

PR #797, opened 2017-12-02
chore(readme): adding bozdoz

afshinm:


afshinm:
Is that correct @bozdoz ?

bozdoz:
Yeah; looks good, thanks! :)

PR #801, opened 2017-12-12
Scroll parent element to targetElement; resolves #210; resolves #350; resolves #373

### LINKED ISSUE #210: Tour tooltip position inside elements with scroll is wrong
The way the offsets are calculated doesnt work if the element is inside another element with scroll. I fixed it changing _getOffset function and using "getBoundingClientRect" function (more reliable than the while loop approach)

```javascript
function _getOffset(element) {
    var elementPosition = {};

    //set width
    elementPosition.width = element.offsetWidth;

    //set height
    elementPosition.height = element.offsetHeight;

    //calculate element top and left
    /* ( Comment old way to get the offsets )
    var _x = 0;
    var _y = 0;
    while (element && !isNaN(element.offsetLeft) && !isNaN(element.offsetTop)) {
        _x += element.offsetLeft;
        _y += element.offsetTop;
        element = element.offsetParent;
    }
    //set top
    elementPosition.top = _y;
    //set left
    elementPosition.left = _x;
    */

    var clientRect = element.getBoundingClientRect();
    elementPosition.top = clientRect.top;
    elementPosition.left = clientRect.left;

    return elementPosition;
}
```

### LINKED ISSUE #350: Problem with div inside/below scrollable div
When IntroJs is used on an element inside a scrollable div, it looks like the position is calculated wrongly. See fiddle here:

http://jsfiddle.net/s9ypgo6c/1/

### LINKED ISSUE #373: Scroll inside child div of body
I don't know if I use the plugin correctly, but I have tried it in several different pages with the same result.

The problem:

I have an absolute header and footer, and an absolute main "DIV". Inside this main DIV, I have a smaller one with overflow: auto and bigger height that the parent DIV, so I can have a scrollbar. 

The problem is that the plugin doesn't scroll inside this small scrollbar, but instead it keeps scrolling the body! So the end result is very bad.

Do I do something the wrong way, is this a known issue?

Thanks :)

bozdoz:
I added another example and tested several of the pages at different screen sizes and scroll positions.  Let me know if this works.  Thanks! 🕺

### DIFF
--- a/example/hello-world/withScroll.html
+++ b/example/hello-world/withScroll.html
@@ -1,0 +1,99 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Basic usage</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="description" content="Intro.js - Better introductions for websites and features with a step-by-step guide for your projects.">
+    <meta name="author" content="Afshin Mehrabani (@afshinmeh) in usabli.ca group">
+
+    <!-- styles -->
+    <link href="../assets/css/bootstrap.min.css" rel="stylesheet">
+    <link href="../assets/css/demo.css" rel="stylesheet">
+
+    <!-- Add IntroJs styles -->
+    <link href="../../introjs.css" rel="stylesheet">
+
+    <link href="../assets/css/bootstrap-responsive.min.css" rel="stylesheet">
+
+    <style>
+      .marketing {
+        height: 300px;
+        height: 40vh;
+        margin-top: 500px;
+        margin-top: 40vh;
+        overflow: auto;
+      }
+    </style>
+  </head>
+
+  <body>
+
+    <div class="container-narrow">
+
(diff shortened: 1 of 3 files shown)


EhsanKia:
Looks much better to me. Using getBoundingClientRect is far cleaner than the current hacky solution.

EhsanKia:
Any updates on this? No only does it fix many issues with sites that have stranges positioning (getBoundingClientRect is far more reliable and consistent), it also makes the code cleaner and also adds extra tests too.

bozdoz:
Just want to make sure @afshinm does a review.

thiagokernell:
I have the same problem, but using a modal. Someone managed to solve in a modal?
6177816a-822d-11e4-97cf-2d7ad9e8f60f

vishwasracharya:
I am experiencing an issue with Intro.js version ^7.2.0. Has this been addressed, or is there an alternative solution available? Could someone please provide information on this matter? If the issue has been resolved, will I need to make any changes to my code to ensure it works properly?

PR #802, opened 2017-12-13
Replaces z-index manipulation with aperture overlay

levkazar:
I'm using intro.js in my SAP Fiori projects. The complex stacking contexts of my DOM would not allow intro.js to raise even the simplest labels above the overlay. So I wrote a wrapper for my projects, which
- sets the opacity of the original overlay to 0,
- and adds four new overlay divs around the highlighted element, like an aperture.   

I propose to take these changes over into the standard. This change has the benefit of removing any dependency on z-index manipulaton of the original DOM.

Detailed changes:
- Removed all z-index manipulation of the original DOM.
- Replaced the global overlay with four overlays around the highlighted element. 
- Had to remove the color gradient on the overlay, because the gradients on the four seperate overlays do not match.
- Had to set overflow: hidden on the body to remove scroll bars. Scroll bars complicate the placement of the seperate overlays.
- Moved auto-scrolling before layer placement to get correct bounding rectangles.

Example with colored overlays for illustration:
image


bozdoz:
I have no idea how this PR got so messed up.

PR #803, opened 2017-12-20
reliable events and instances

### LINKED ISSUE #433: Does not release keypress listeners on exit
Please see this sample:
https://jsfiddle.net/xe9vh35f/

If `introJs().exit()` is called, the tooltips are dismissed, but if a user hits "enter" or the left/right arrow keys, the intro tooltips continue to be opened (sidenote: there is no overlay when they are open).

bozdoz:
added DOMEvent helper for events; _stamp helper for uniquely identifying objects; added instances method for getting a specific introJs instance; resolves #433 as best as anything could;

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -2273,7 +2273,7 @@
     // add instance to list of _instances
     // passing group to _stamp to increment
     // from 0 onward somewhat reliably
-    introJs[ _stamp(instance, 'introjs-instance') ] = instance;
+    introJs.instances[ _stamp(instance, 'introjs-instance') ] = instance;
 
     return instance;
   };
--- a/intro.js
+++ b/intro.js
@@ -112,8 +112,7 @@
    * @returns {Boolean} Success or not?
    */
   function _introForElement(targetElm) {
-    var introItems = [],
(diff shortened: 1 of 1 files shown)


intro.js
```diff
@@ -247,92 +246,87 @@
...
-          if(e.preventDefault) {
-            e.preventDefault();
-          } else {
-            e.returnValue = false;
-          }
-        }
-      };
+  function _onResize () {
+    this.refresh.call(this);
+  }
```
bozdoz: seems redundant

intro.js
```diff
@@ -1472,6 +1463,112 @@
...
+      *
+      * @param Object obj
+      * @param String type        event type
+      * @param Function listener
+      * @param Object context
+      * @return String
+      */
+      this._id = function (obj, type, listener, context) {
+        return type + _stamp(listener) + (context ? '_' + _stamp(context) : '');
+      };
```
bozdoz: The id guarantees that we can reliably add and remove the same event listener. 

Note that `.bind` creates a new function.

PR #807, opened 2018-01-09
Fix tooltip and highlight off problem due to scrolling

sophia-spoke:
We encountered an issue where if an element is scrolled up, the tooltip and helper layer placement become incorrect. Turns out this is because `_getOffset` method didn't take the scrolling into account when looking at and calculating the offsetTop. 

Problem screenshot: 
screen shot 2018-01-08 at 8 03 54 pm


bozdoz:
This should be resolved in #801

PR #812, opened 2018-01-23
Event listeners with promises === more scripting power for developers

levkazar:
There are situations where developers need to interrupt an intro's flow, e.g. to wait for a page transition to finish or a modal to render, before the next step can be shown. This PR introduces event handlers that can return promises. The promise delays the program flow until it resolves or rejects, which enables developers to finish any asynchronous task before the intro proceeds. 

Features:
- Event handlers can return any value, but only Boolean, Promise and undefined (i.e. no return statement) are processed.
- Event names are
  - Start (new, fires right before the first tooltip is shown)
  - BeforeChange (can return false to stop displaying the tooltip)
  - Change
  - AfterChange
  - Complete
  - BeforeExit (can return false to prevent exiting)
  - Exit
  - HintsAdded
  - HintClick
  - HintClose
- Introduces new methods for event handler registration. Below, 'Event' in the function signature is an event name like 'Change', and all functions are written lower camel-case:
  - `attachEvent(callbackFunction, callbackContext)`: Registers a callback function the context it should be called with. callbackContext is optional.
  - `attachOnceEvent(callbackFunction, callbackContext)`: Same as attachEvent, but the callback is excuted only once, and then automatically detached.
  - `detachEvent(callbackFunction, callbackContext)`: Removes an event handler from the callback list. Possible only if callbackFunction and callbackContext are the same references that where used when registering.
- The `attachEvent` methods return a function that can be called to deregister the event handler. This is an alternative to `detachEvent`.
- For each event, any number of listener functions can be registered. This breaks the limitation of a single callback function.
- The `onevent` functions still exist as an alternative to `attachEvent`. They are written all-lower-case. This ensures compatibility with the previous event handling framework.
- If a promise from an event handler rejects, then the intro flow is stopped.

An example is given in folder `example/multi-page-promises`.

Basic structure of event handlers:
```
intro.attachBeforeChange(beforeChange, this);

function beforeChange() {
  return new Promise(function(resolve, reject) {
      // do some asynchronous task here, then resolve or reject.
      // The event BeforeChange can resolve with a boolean, so resolve(true|false) is possible.
      // When calling reject(), the intro will stop immediately
  });
}
```

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -38,6 +38,7 @@
         <li><a href="hint/index.html" title='Basic hints usage'>Basic hints usage</a></li>
         <li><a href="hint/withElement.html" title='Basic hints usage on an element'>Basic hints usage on an element</a></li>
         <li><a href="programmatic/hint.html" title='Programmatic defining hints using JSON'>Programmatic defining hints using JSON</a></li>
+        <li><a href="multi-page-promises/index.html" title='Event handling with Promises'>Event handling with Promises</a></li>
       </ul>
     </div>
   </body>
--- a/example/multi-page-promises/index.html
+++ b/example/multi-page-promises/index.html
@@ -1,0 +1,183 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <title>Multi-page introduction, Page 1</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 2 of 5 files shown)


Avejack:
This would be the best feature for this library! Needed. Is there an ETA?

amiceli:
Any news ?

PR #818, opened 2018-02-02
Add callbacks for close intro on overlay layer click and esc

Rxbsxn:
I have added some quite important callbacks when we want to handle different actions in different intro close case.

### DIFF
--- a/example/callbacks/onescapeexit.html
+++ b/example/callbacks/onescapeexit.html
@@ -1,0 +1,81 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Callbacks: onoverlayclickexit</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="description" content="Intro.js - Better introductions for websites and features with a step-by-step guide for your projects.">
+    <meta name="author" content="Afshin Mehrabani (@afshinmeh) in usabli.ca group">
+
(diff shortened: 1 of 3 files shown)


bozdoz:
I'm not sure of a use case for this. In any case, it could work.  I might prefer something more generic; something event-based, instead of callback-based, like:

```javascript
intro.on('exit', function () {
  console.log('something fired exit');
});
```

where the code in Intro.js is replaced with something like:

```javascript
this.fire('exit');
```

Rxbsxn:
@bozdoz Could you explain your approach a little more because you use callbacks everywhere in code?

bozdoz:
> "explain your approach"

The callback style of development requires (1) checking for a function after an event, then (2) executing it.  An event-based approach would only require firing an event, where the event handling deals with executing **any number** of functions (note that the callback approach would only execute one).  Also, the way this is currently implemented, there's also this strange idea of creating a callback function with a method:

```javascript 
onclick: function (cb) {
  this._onclick = cb;
} 
```

But, this could easily be done directly like this:

```javacript
intro._onclick = function () {};
```

All I'm saying is that this library ought to be cleaned up a bit.  **I have no objection to your Pull Request**.  I have only hesitation, because I think the library ought to be optimized.  For what it's worth, I added a recommendation to move towards event-based logic in #602.

dcantatore:
@afshinm - Are any of the pull requests going to be merged in?

PR #824, opened 2018-02-20
Avoid Content Security Policy violation

danbower:
Hey,

Firstly I'd just like to say thanks for taking the time to create such a wonderful library. I've had to make some slight modifications as we have Content Security Policy enabled with pretty strict policies and unfortunately intro.js currently violates one of them. I thought this change might be useful for other people too. Here's a bit more detail from my commit message:

> The setAttribute('style', foo) usage was violating the "child-src" CSP
> directive when the directive source is set to "self". This is
> functionally equivalent but avoids the violation.

Cheers,
Dan

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -954,10 +954,10 @@
       }
 
       //set new position to helper layer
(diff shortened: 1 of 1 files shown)


bozdoz:
Cool! Thanks!

PR #826, opened 2018-02-23
Aperture background as solution to stacking context problem

bozdoz:
This PR is a remake of #802 because that PR somehow got messed up.  I recreated it as best I could, and added a test HTML page (which works!).  One problem is that the CSS transitions cause the backdrop to do funny things (understandably).  Otherwise, this is a great solution!  Take a look!


bozdoz:
Just thinking about this some more.  Probably we could rethink the logic for the aperture.  Instead of this:

screenshot from 2018-02-25 14 00 01

We could do this (more like a real camera aperture):
screenshot from 2018-02-25 14 01 31

Doing it this way should be easier (and make for less animation).  In the first example, the top part, above the highlighted element, has three columns, which each animate `left` and `width`: top-left animates a new `width`, top-center animates a new `left` and `width`, and top-right animates a new `width`.  In the **second example**, the top part is made of two columns, one animates `width`, and the other animates `left` and `width`.  So that cuts out one of the animations, and *might* reduce the likelihood of a glitch appearing.  

Other ways I can see this working is by having a parent element with `position: absolute`, and these inner moving parts as `position: relative` where only the left-most parts are recalculated, and the right-most automatically fill the remaining space in the browser window.

DanielSchuech:
I have tested this pr and can confirm that it also works in more complex situations. Like you mantioned the backdrop is a bit buggy. If you can fix that it would work like a charm. I would like to see this in a further release.

levkazar:
After some experiments with the aperture overlay I came to the conclusion that a smooth transition is only possible by controlling the animation loop, using requestAnimationFrame. The animation properties have to removed from the css to make this work properly. 

You can check out an early working version here:
https://github.com/levkazar/intro.js/tree/smooth-animation

In our productive code we are now using a heavily modified version of intro.js. Maybe I will find the time to make these modifications into a PR.

bozdoz:
The real camera aperture idea seems to have worked!  I noticed some minor glitches when resizing the page back and forth rapidly, but I think that is an extremely minor case.  I'm still not sure how we're dealing with scrolling pages.  That buggyness remains. I think I'll have to check master again to see how that's filling the screen with the overlay.  I also have not (nor have I ever) tested floating positioned elements.  Also, I have not tested my second idea of calculating the left-most widths only, and having the right-most parts fill the remaining width.  I think that would completely eliminate the rapid resizing glitch.

One more also: I'm still unsure if this PR is globally useful; but, I have a feeling that it's a far better approach, so hopefully we can figure out all the potential bugs along the way.

Behold (I put the colors back on for reference):

screenshot from 2018-02-27 12 15 29

levkazar:
Great! Please make sure to test with large pages. I encountered the worst glitches with big page heights.

bozdoz:
Also thinking maybe the window resizing glitch has to do with repeatedly firing the 'resize' event.  Perhaps we can throttle it and the glitch will go away.  Also, I just tested the second idea: to auto-fill the remaining width of the latter two parts.  Check it out on codepen: https://codepen.io/bozdoz/pen/qxQGJr?editors=0110

bozdoz:
Actually, you know what: it could be easier just having the overlays overlapping now that they're in a container (see v2): https://codepen.io/bozdoz/pen/VQqvmd?editors=0110

For this one, only 4 variables are calculated! Having a container means that we can assign the opacity to the container, so that the overlapping children don't appear to overlap.  This seems far better.

bozdoz:
And another with scroll: https://codepen.io/bozdoz/pen/eVbZPm?editors=0110

I think we should disable scrolling when the element is in focus.  Not sure why anyone would want to scroll on a page where the black backdrop is covering anything else.  Thoughts?

DanielSchuech:
On large pages it could be very usefull for users to be able to scroll the page. 

example: singleside application with big height
You are creating a tutorial and focues at the end an element at the bottom of the page. A user just want to have a look to some details on the top of the page to better understand the current step. Then he can scroll up and view the information he needs through the backdrop. 

If you disable scrolling completly the user would have to go all steps back and there could by possibly a lot. That would be a bad user experience in my opinion.

DanielSchuech:
I ve tested your latest changes and the backdrop looks now better. I think you missed the `z-index` property in the `introjs-fixedTooltip` css class. Other fixed elements would be displayed over the backdrop through the stackigng context. I ve changed it to the following and now its working:
```css
.introjs-fixedTooltip {
    position: fixed;
    z-index: 999998;
}
```

bozdoz:
The main problem as I see it seems to be of trying to fill the entire background with the backdrop.  Maybe I can do this programmatically with scrollHeight.  I'm also wondering about showing the intro for a specific element, and not the document body.  Some things to consider, anyway.

alexthewilde:
I can confirm that this solution works very well, especially for targeting fixed positioned elements.

cshapeshifter:
I think this is a great improvement! The existing approach of messing with z-index doesn't work at all for me.

One problem I'm having is that sometimes there are thin, white borders where the 4 different pieces don't quite line up, as seen below.
2019-05-09-033325_1026x834_scrot
I think a simple solution would be to replace the 4 separate divs by a single div for the duration of the animation only.

bozdoz:
This is a possibly much better (unimplemented) solution to the problem: https://codepen.io/bozdoz/pen/VQqvmd?editors=0110

Re: comment above: 
> overlapping children don't appear to overlap

opacity is on the container and the 4 sections just overlap, meaning no white borders @cshapeshifter

cshapeshifter:
@bozdoz Thanks, this solution works very well for me.

afshinm:
Working on a fix here https://github.com/usablica/intro.js/pull/1075 (v3 compatible)

PR #845, opened 2018-04-24
onEnter callback

Vasiliy-Bondarenko:
callback allows to make something on the page when user enters new step.
example: I have several small elements on the page in separate places and I want to blink them to grab user's attention to them.


bozdoz:
Could you provide example code for how to use this? Thanks!

PR #858, opened 2018-05-11
Overlay triggers next step

### LINKED ISSUE #857: Overlay clicked triggers next step
## Description
Overlay clicked will trigger next step or done button when configured. A new option should be added or the `exitOnOverlayClick` refactored to allow for more options on the overlay click target.

## Expected Behavior
Overlay clicked will trigger next step or done button when configured. 

## Actual Behavior
The tour closes or does nothing dependant on the `exitOnOverlayClick` option.

TCastelein:
fixes #857 

## Changes
- New option: `overlayClicked` has 3 possible options:
  - `''` Do nothing (same as `exitOnOverlayClick = false`)
  - `'exit'` is now the default to have everything backwards compatible
  - `'continue'` mimics the next and done buttons when overlay is clicked
- Deprecated option `exitOnOverlayClick` as this is covered by the new option. (Still backwards compatible)
- Some whitespace removal my editer did automaticly (removed trailing whitespaces) if necesary i can return them but the change is clear when [hiding the whitespace in the diff](https://github.com/usablica/intro.js/pull/858/files?utf8=%E2%9C%93&w=1)

This is my first PR here so please inform me if I did something wrong or missed something. I will then update my PR

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -64,8 +64,10 @@
       highlightClass: '',
       /* Close introduction when pressing Escape button? */
       exitOnEsc: true,
-      /* Close introduction when clicking on overlay layer? */
-      exitOnOverlayClick: true,
+      /* Deprecated use overlayClicked Close introduction when clicking on overlay layer? */
+      exitOnOverlayClick: false,
+      /* Action when overlay is clicked. Options are: 'exit' or 'continue' */
+      overlayClicked: 'exit',
       /* Show step numbers in introduction? */
       showStepNumbers: true,
       /* Let user use keyboard to navigate the tour? */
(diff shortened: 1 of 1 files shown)


intro.js
```diff
@@ -1780,7 +1782,19 @@
     targetElm.appendChild(overlayLayer);
 
     overlayLayer.onclick = function() {
-      if (self._options.exitOnOverlayClick === true) {
+      if (self._options.overlayClicked === '' || self._options.exitOnOverlayClick === false) {
+        // Do nothing
```
TCastelein: This bit is needed for the backwards compatibility as both `overlayClicked` and `exitOnOverlayClick` are in play. I have the `'exit'` default as last if so we can use both together. In the furture `exitOnOverlayClick` should be removed and this code made easier to read.

PR #894, opened 2018-10-01
Aperture overlay

chrwolff:
The old idea of an aperture overlay instead of z-index shifts. Not working well for large pages; some jitter.


bozdoz:
Dupe of #826 ?

PR #906, opened 2018-11-22
Single Step error on Right key

TriSysBusinessSoftware:
Where only a single step is in the chain and the user presses the right arrow key, an error is raised because the nextStep object is undefined.
introjs github


afshinm:
Fixed.

PR #927, opened 2019-05-20
Fixed calculating tooltip bottom position

kirillovarchi:
@bozdoz have a look please. It's really small and useful fix for mobile devices with small width resolution.


bozdoz:
Looks good to me

Issue #935, opened 2019-07-02
disableInteraction in options not working


DeveloperSlope3:
## Description
I want all steps disableInteraction, so I put them in the options of introJS, not as an atribute of the DOM element.

## Expected Behavior
I expect all the steps have the interaction disabled

## Actual Behavior
Nothing happens, the interaction is enabled ignoring the value I set. (as an atribute it works perfectly)

JohnArcher:
Works for me. Maybe you should show your code.

MrViolin:
The source code, line 186:

> if (typeof (currentElement.getAttribute('data-disable-interaction')) !== 'undefined') {
>   disableInteraction = !!currentElement.getAttribute('data-disable-interaction');
>  } else {
>   disableInteraction = this._options.disableInteraction;
>  }

DOM.getAttribute(attr), It's type will never be undefined. It will be 'null' or 'string'
You can change it like this:

> if(!currentElement.getAttribute('data-disable-interaction'))

now, the global options 'disableInteraction' can works.

MrViolin:
The source code, line 186:

> if (typeof (currentElement.getAttribute('data-disable-interaction')) !== 'undefined') {
>   disableInteraction = !!currentElement.getAttribute('data-disable-interaction');
>  } else {
>   disableInteraction = this._options.disableInteraction;
>  }

DOM.getAttribute, It's type will never be undefined.
You can change it like this:

> if(currentElement.getAttribute('data-disable-interaction'))

now it works.

PR #942, opened 2019-08-17
Fix disableInteraction option

m417z:
Fixes issue https://github.com/usablica/intro.js/issues/935

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -183,7 +183,7 @@
 
         var step = parseInt(currentElement.getAttribute('data-step'), 10);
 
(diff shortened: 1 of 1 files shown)


m417z:
Here's some activity.
It's a simple patch to a simple issue, I see no reason not to merge it. I'm using a fork with the fix and it works property.

m417z:
Here's some activity.

afshinm:
Hey @m417z, thanks. I'm going to fix the branch conflicts and merge into master.

PR #953, opened 2019-11-22
Fix scolling into scrollable parent div

shaun86wang:
The function _scrollParentToElement only handles situation if element.offsetParent is the parent. If the element is deeply nested inside the scrollable parent, the parent then would not scroll enough to show the specified element (https://codepen.io/shaun86wang-the-bashful/pen/dyyLmNG). The changeset fixed that issue

### DIFF
--- a/intro.js
+++ b/intro.js
@@ -2279,7 +2279,7 @@
   */
   function _scrollParentToElement (parent, element) {
     var offsetTop = 0;
-    for (var e = element; e != parent; e = e.offsetParent) {
+    for (var e = element; e !== parent; e = e.offsetParent) {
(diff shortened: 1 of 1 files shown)


javierdiegof:
I'm experiencing the same problem. When trying to calculate the amount to scroll inside the scrollable parent: 
`
function _scrollParentToElement (parent, element) {
    parent.scrollTop = element.offsetTop - parent.offsetTop;
  }
`
element.offsetTop returns the offset between the element trying to be focused and its parent (the offset parent), which is not the scrollable parent since the element is nested (the offset parent is in-between).

PR #1001, opened 2020-04-14
Fix for method being called on null error

jbaseas:
Fixes error on cases of intro with only a single slide.


afshinm:
Manually added.

PR #1024, opened 2020-10-11
chore: updating the release script to conform to the v3 format

afshinm:
@jalalazimi this is not the best approach but it works. I basically tried to update the `release.sh` script to match the v2 format. This way, we can keep the ./dist folder as well. 

Let me know what you think.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -6,14 +6,13 @@ import progress from 'rollup-plugin-progress';
 import json from '@rollup/plugin-json';
 import pkg from './package.json';
 import { terser } from 'rollup-plugin-terser';
-import sass from 'rollup-plugin-sass';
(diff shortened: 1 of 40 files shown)


rollup.config.js
```diff
@@ -29,7 +30,7 @@ const banner = `/*!
 const plugins = [
   json(),
   sass({
-    output: "dist/introjs.css",
+    output: `${outputPath}/minified/introjs.min.css`,
```
jalalazimi, line 33: Why do we need to keep the minified files in a separate folder?

afshinm: to match the existing package structure, see this https://www.jsdelivr.com/package/npm/intro.js

jalalazimi (approved):

PR #1047, opened 2020-10-17
Add additional class option for progressbar

LeonBeilis:
Best practice to allow developers assign their theme classes to the progress bar without write weird workarounds,
I needed this and didn't found existing solution.


LeonBeilis:
@afshinm would like to hear your opinion about this

afshinm:
I like this idea but I believe we should add a different option to add a custom CSS class to the root Introjs element (instead of appending custom CSS classes to each element).

afshinm:
I like it @LeonBeilis, thanks. I will release a new version today.

PR #1088, opened 2020-11-26
feat: adding `title` field to intro items

afshinm:


bozdoz:
A note: Are we using yarn or npm? Using yarn it prints this warning:

>  package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.

afshinm:
@bozdoz yeah, looks like we have two locks at the moment and we can only keep the npm lock (package-lock.json) in that case? I can update it after merging this PR.

afshinm:
just pushed my final changes @bozdoz

src/core/placeTooltip.js
```diff
@@ -197,11 +195,6 @@ export default function placeTooltip(
 
   arrowLayer.style.display = "inherit";
 
-  if (typeof helperNumberLayer !== "undefined" && helperNumberLayer !== null) {
```
bozdoz, line 200: Why is this no longer needed?

src/index.js
```diff
@@ -33,17 +33,19 @@ function IntroJs(obj) {
...
   this._options = {
     /* Next button label in tooltip box */
-    nextLabel: "Next &rarr;",
+    nextLabel: "Next",
     /* Previous button label in tooltip box */
-    prevLabel: "&larr; Back",
+    prevLabel: "Back",
     /* Skip button label in tooltip box */
-    skipLabel: "Skip",
+    skipLabel: "×",
```
bozdoz, line 40: Probably bad for accessibility. Suppose it may need to be provided for translations to though.

src/styles/introjs.scss
```diff
@@ -202,56 +191,100 @@ tr.introjs-showElement {
...
+  line-height: 32px;
+}
+
+.introjs-tooltip-header {
+  padding-left: 20px;
+  padding-right: 20px;
+  padding-top: 10px;
+
+  &:after {
+    content: ".";
```
bozdoz, line 221: Does this need a period?

src/styles/introjs.scss
```diff
@@ -264,24 +297,31 @@ tr.introjs-showElement {
...
+  &:hover {
+    color: $black900;
+    outline: none;
+    text-decoration: none;
+  }
 
-.introjs-prevbutton.introjs-fullbutton {
-  border: 1px solid $black400;
-  border-radius: 0.2em;
+  &:focus {
```
bozdoz, line 312: &:hover,
&:focus {
?

src/styles/introjs.scss
```diff
@@ -320,11 +360,13 @@ tr.introjs-showElement {
 
 .introjs-bullets {
   text-align: center;
+  padding-top: 10px;
+  padding-bottom: 10px;
```
bozdoz, line 364: Just a thought: wonder if this would be worth converting into a $spacing variable. Could be easier to manage/extend. Could use default too so it's easily overridden:

$spacing: 10px !default;

bozdoz (commented):
Just some questions and thoughts

bozdoz (changes requested):
Close button needs some width/padding if it's just an 'x'. The touch target is too small. Material design recommends 48px.

src/core/placeTooltip.js
afshinm: @bozdoz I changed the position of that element and we don't need to adjust it after every refresh. It's in the tooltip container now.

src/styles/introjs.scss
afshinm: good idea. I can work on it in a different PR.

src/index.js
afshinm: correct

bozdoz:
Looks good to me

bozdoz (approved):

PR #1090, opened 2020-11-28
Hot Reloading Examples page

bozdoz:
adds: 
- browser-sync as a hot reloading dev server
- npm-run-all for running browser-sync and rollup in parallel

### DIFF
--- a/bs-config.js
+++ b/bs-config.js
@@ -1,0 +1,21 @@
+/*
+ | Browser-sync config file
+ | For up-to-date information about the options:
+ |   http://www.browsersync.io/docs/options/
+ | */
+ module.exports = {
+  "ui": false,
+  "files": ['dist/**', 'example/**', 'docs/**', 'themes/**'],
+  "watchEvents": [
+      "change"
+  ],
+  "ignore": ["node_modules/*"],
+  "single": false,
+  "watchOptions": {
+      "ignoreInitial": true
+  },
+  "server": true,
+  "port": 4000,
+  "notify": false,
+  "startPath": "/example"
+};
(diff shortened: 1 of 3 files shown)


bs-config.js
```diff
@@ -0,0 +1,21 @@
+/*
+ | Browser-sync config file
+ | For up-to-date information about the options:
+ |   http://www.browsersync.io/docs/options/
+ | */
+ module.exports = {
+  "ui": false,
+  "files": ['dist/**', 'example/**', 'docs/**', 'themes/**'],
```
afshinm, line 8: I'm planning on moving the examples/ folder to introjs-website repo. I think we should keep this repo for Intro.js codebase and the `introjs-website` for docs, examples, etc.

bozdoz: Ah, I was wondering about that.  That makes this whole PR dead.  I can make new instructions once those changes are made I guess.  We could run rollup -c -w in this repo, and then browsersync in the website, with yarn link or something?  Cool to close this PR?

bozdoz: Actually what happens with the netlify deploys if the example dir is moved?

afshinm: That's a good question. I did not consider the Netilfy deploys :thinking:

afshinm: git sub-modules? :fearful:

bozdoz: Not sure if there are any good solutions.  Symlinking and string replacements (for URL's) come to my mind.  I guess it's a good question as to what the examples pages are for.  Are they for the website? or for development?

afshinm: both? I thing the end goal is to have the examples folder on the website but it's also nice to have that folder in the main repo (helps with Netlify deployments as you mentioned).

PR #1091, opened 2020-11-28
Custom Event handling methods as solution to circular dependencies

bozdoz:
Add events to deal with circular dependencies; added some type annotations and definitions; cleaned up some methods.

Also, events could make for a nicer/more flexible API, going forward.

Think of this line:

```js
  // signal to all listeners that introjs has exited
  this.fire('exit');

  // check if any callback is defined
  if (this._introExitCallback !== undefined) {
    this._introExitCallback.call(this);
  }
```

Devs can add all the introJs.on('exit') handlers they want.

### DIFF
--- a/src/core/exitIntro.js
+++ b/src/core/exitIntro.js
@@ -1,7 +1,4 @@
 import forEach from "../util/forEach";
-import DOMEvent from "./DOMEvent";
-import onKeyDown from "./onKeyDown";
-import onResize from "./onResize";
 import removeShowElement from "./removeShowElement";
 import removeChild from "../util/removeChild";
 
@@ -9,6 +6,7 @@ import removeChild from "../util/removeChild";
  * Exit from intro
  *
  * @api private
+ * @this {import('./IntroJs').default}
  * @method _exitIntro
  * @param {Object} targetElement
  * @param {Boolean} force - Setting to `true` will skip the result of beforeExit callback
@@ -34,15 +32,6 @@ export default function exitIntro(targetElement, force) {
     forEach(overlayLayers, (overlayLayer) => removeChild(overlayLayer));
   }
 
-  //remove all helper layers
-  const helperLayer = targetElement.querySelector(".introjs-helperLayer");
-  removeChild(helperLayer, true);
-
-  const referenceLayer = targetElement.querySelector(
-    ".introjs-tooltipReferenceLayer"
-  );
-  removeChild(referenceLayer);
-
   //remove disableInteractionLayer
   const disableInteractionLayer = targetElement.querySelector(
     ".introjs-disableInteraction"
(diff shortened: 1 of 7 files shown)


src/core/exitIntro.js
```diff
@@ -34,15 +32,6 @@ export default function exitIntro(targetElement, force) {
     forEach(overlayLayers, (overlayLayer) => removeChild(overlayLayer));
   }
 
-  //remove all helper layers
-  const helperLayer = targetElement.querySelector(".introjs-helperLayer");
```
afshinm, line 38: Why we don't need this anymore?

src/core/exitIntro.js
```diff
@@ -55,15 +44,13 @@ export default function exitIntro(targetElement, force) {
 
   removeShowElement();
 
-  //clean listeners
-  DOMEvent.off(window, "keydown", onKeyDown, this, true);
-  DOMEvent.off(window, "resize", onResize, this, true);
```
afshinm, line 60: Are we handling these somewhere else?

src/core/exitIntro.js
```diff
@@ -55,15 +44,13 @@ export default function exitIntro(targetElement, force) {
...
+  // signal to all listeners that introjs has exited
+  this.fire('exit');
```
afshinm, line 48: :100:

src/core/exitIntro.js
afshinm: Oh nvm, just saw the `off` handler. nice

afshinm (commented):
I really like this idea but I believe we should still support the `introjs(...).oncomplete(fn => ...)` notion as well (in addition to `introjs(...).of('complete', fn => ...)`. I don't really want to introduce a massive backward incompatibility. But I can definitely see that this approach can simplify the codebase.

src/core/exitIntro.js
bozdoz: removed in another file

bozdoz:
> I really like this idea but I believe we should still support the `introjs(...).oncomplete(fn => ...)` notion as well (in addition to `introjs(...).of('complete', fn => ...)`. I don't really want to introduce a massive backward incompatibility. But I can definitely see that this approach can simplify the codebase.

I'll take a look at `oncomplete`.  Did I remove any of that functionality?

bozdoz:
Just noticing that exitIntro is always called after the oncomplete:

```js
      if (
        this._introItems.length - 1 === this._currentStep &&
        typeof this._introCompleteCallback === "function"
      ) {
        this._introCompleteCallback.call(this);
      }

      exitIntro.call(this, this._targetElement);
```

maybe we can just put that at the top of exitIntro instead

afshinm:
> I'll take a look at oncomplete. Did I remove any of that functionality?

Sorry no, I sort of assumed the long term plan is to replace the new event handle (`intro.on(...)`) with the callbacks (`intro.on...()`)?

> maybe we can just put that at the top of exitIntro instead

That makes sense @bozdoz.

bozdoz:
> Just noticing that exitIntro is always called after the oncomplete:
> 
> ```js
>       if (
>         this._introItems.length - 1 === this._currentStep &&
>         typeof this._introCompleteCallback === "function"
>       ) {
>         this._introCompleteCallback.call(this);
>       }
> 
>       exitIntro.call(this, this._targetElement);
> ```
> 
> maybe we can just put that at the top of exitIntro instead

I was wrong about this. It's called if skip button is pressed on the last tip, and if nextStep is called when there is no nextStep.  Seems fair to keep it.

PR #1117, opened 2020-12-13
fix(onKeyDown): adding NumpadEnter

afshinm:


bozdoz:
Surprised to see that; pressing the keypad for me just says 'Enter'.  🤷‍♀️

bozdoz (approved):

afshinm:
yep! but looks like there actually an event for that https://www.w3.org/TR/uievents-code/

PR #1118, opened 2020-12-15
fix(showElement): next/done button needs to be displayed when tour has one step

### LINKED ISSUE #1101: 3.1.0 don't show done/next button in only one step
I have tried all possible combinations of "show buttons", "nextToDone", etc ...
And when there is only one step, it does not show any type of button.

```
var recuerdaAgentes = introJs();
    recuerdaAgentes.setOptions({
        steps: [
        {
        element: '#agente1',
        intro: 'Sr. PRIMER APELLIDO, le recordamos que nuestro personal disponible puede <b>atenderle en vivo</b> cuando lo necesite durante el horario de atención:<br>9:00h - 19:00h'
        }
        ],
        nextLabel: 'De acuerdo, gracias',
        doneLabel: 'De acuerdo, gracias',
        overlayOpacity: .8,
        exitOnOverlayClick: false,
        exitOnEsc: false,
        keyboardNavigation: false,
        nextToDone: true,
        showButtons: true,
        showBullets: false,
        showProgress: false,
        showStepNumbers: false,
        disableInteraction: true

    });
    recuerdaAgentes.start();
```

HTML RESULT: https://jmp.sh/lQNnE2V
SCREEN RESULT: https://jmp.sh/vXeDyX1

afshinm:
Fixes #1101

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -9,6 +9,7 @@
     "exports": true,
     "define": true,
     "SVGElement": true,
+    "Element": true,
     "module": true,
     "console": true,
     "global": true,
--- a/.jshintrc
+++ b/.jshintrc
@@ -5,10 +5,6 @@
   "eqeqeq": true,
   "globals": {
     "window": true,
-    "jest": true,
-    "describe": true,
-    "expect": true,
-    "test": true,
     "document": true,
     "exports": true,
     "define": true,
@@ -16,6 +12,24 @@
     "module": true,
     "console": true,
     "global": true,
-    "self": true
+    "self": true,
+
+    "afterAll": false,
+    "afterEach": false,
+    "beforeAll": false,
+    "beforeEach": false,
+    "check": false,
+    "describe": false,
+    "expect": false,
+    "gen": false,
+    "it": false,
+    "fit": false,
+    "jest": false,
+    "pit": false,
+    "require": false,
+    "test": false,
+    "xdescribe": false,
+    "xit": false,
+    "xtest": false
   }
 }
--- a/tests/util/addClass.test.js
+++ b/tests/util/addClass.test.js
@@ -51,4 +51,12 @@ describe("addClass", () => {
 
     expect(el.className).toBe("firstClass");
   });
+
+  test("should not append duplicate list of classNames to an empty className", () => {
+    const el = document.createElement("div");
+
+    addClass(el, "firstClass firstClass firstClass");
+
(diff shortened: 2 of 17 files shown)


tests/util/addClass.test.js
```diff
@@ -24,5 +24,32 @@ describe('addClass', () => {
 
       expect(el.className).toBe('firstClass secondClass');
     });
+
+    test('should not append duplicate classNames to svg elements', () => {
+      const el = document.createElement('svg');
```
bozdoz: `document.createElementNS('http://www.w3.org/2000/svg', 'svg');`

.jshintrc
```diff
@@ -5,17 +5,32 @@
   "eqeqeq": true,
```
bozdoz, line 5: eslint is a dream compared to jshint 😎

tests/util/addClass.test.js
```diff
@@ -1,28 +1,54 @@
...
+    expect(el.className).toBe("firstClass");
+  });
+
+  test("should not append duplicate list of classNames to elements", () => {
+    const el = document.createElement("div");
+    el.className = "firstClass firstClass";
+
+    addClass(el, "firstClass firstClass firstClass");
+
+    expect(el.className).toBe("firstClass");
```
bozdoz, line 52: a very quirky test; I expect this isn't always true.  Only true when `elem.classList !== undefined`; otherwise:

```js
expect(el.className).toBe('firstClass firstClass firstClass firstClass firstClass');
```

but, probably not important. 😆

tests/util/addClass.test.js
afshinm: nice catch

tests/util/addClass.test.js
afshinm: just added a new test for empty classname

bozdoz: Probably not important; I was just saying that for (I guess) IE9 the quirky test case would be:

```javascript
test("does append duplicate list of classNames if element does not support classList", () => {
    const el = { className: '' }; // fake pseudo-element
    const classes = "firstClass firstClass firstClass";

    addClass(el, classes);

    expect(el.className).toBe(" " + classes);
  });
```

Again, doesn't really matter.  Might be worth removing that portion of addClass, since I'm not aware of anyone using IE9 anymore.

bozdoz (approved):

bozdoz:
Great to see so many tests!

Issue #1147, opened 2021-01-22
Positioning breaks when passing relatively-positioned element to introJs() to limit the presentation section 


bendavis78:
If I use introJs on a relatively positioned section (or has a relatively positioned parent), the positioning of the helper layer will be incorrect:

```javascript
introJs('.my-relative-el').start()
```

This is because positioning is determined using `getBoundingClientRect()`, which is relative to the viewport. However, the top/left properties are relative to the parent target element because the helper layer is appended to that target element.

afshinm:
Thanks @bendavis78 -- can you send us an online example (jsfiddle, codesandbox, etc) please?

bendavis78:
Sorry for the late reply, here is a jsfiddle demonstrating the issue:

https://jsfiddle.net/opkjfvw4/27/

rr222cy:
Just installed intro.js into a new project using a sidebar navigation menu and experiencing the same issue due to relatively-positioning. Would be great if @bendavis78 fix could be accepted :) @afshinm

fantypants:
is this the same as https://github.com/usablica/intro.js/issues/855 ?

bendavis78:
@fantypants possibly -- you might want to check out this branch to see if it fixes your issue: https://github.com/BoldIdeaInc/intro.js/commits/issue-1147

bendavis78:
@afshinm can you remove the "question label"? Were you able to see the jsfiddle or review the PR?

PR #1163, opened 2021-02-09
Fixes issue #1147: Positioning breaks when passing relatively-positioned element to introJs()

bendavis78:
This fixes issue #1147. The solution here is to subtract the top/left values of `this._targetElement` from the calculated position.


afshinm:
Thanks @bendavis78

afshinm:
I just tested this with a container that doesn't have relative position, and looks like it breaks a couple of things. Working on a fix.

PR #1190, opened 2021-03-12
chore: update build file to create 2 versions (es5 and raw one)

miloszsobczak:
Not everyone wants to use default babelified version of the package due to multiple reasons:
- modern browser support
- willingness to use smaller version of the file (in this case twice smaller than original one)
- existence of potential conflict with already embedded core-js library on the page

So I decided to push some change that will be still compatible with current version (as main file in package.json has been changed too). If someone wants to use raw code wirtten in UMD style I believe it can be truly beneficial for wider audience.

Let me know what do you think.
The code itself it is not super sophisticated, it just the matter of rollup configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "type": "git",
     "url": "https://github.com/usablica/intro.js"
   },
-  "main": "intro.js",
+  "main": "intro.es5.js",
   "scripts": {
     "prettier": "prettier --write 'src/**/*'",
     "test": "npm run test:jest && npm run test:jshint",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "type": "git",
     "url": "https://github.com/usablica/intro.js"
   },
-  "main": "intro.js",
+  "main": "intro.es5.js",
```
afshinm, line 19: I'm not sure if changing the default main entrypoint is a good idea right now. Not entirely sure how this would affect people. Could you please revert this? You should still have access to the `intro.es5.js` in the npm package/CDNs.

afshinm (changes requested):
Sounds good.

PR #1229, opened 2021-05-05
fix bug (checkRight)

JiaZ-star:
When the "tooltipPosition" is set to "right", the configuration not in effect.

So I modified the "checkRight.js" file


afshinm:
Thanks @JiaZ-star. I'm not sure if I understand the issue. Could you please explain the issue that you're seeing? And how this can fix it? It also would be great if you send an online example (jsfiddle, codesandbox)

Issue #1202, opened 2021-04-06
Interaction not working properly when multiple stacking contexts exist.


KevinMoran99:
## Description
I'm working on a project in which I'm trying to implement introjs, but I've noted that when highlighting an element that's inside of a different stacking context (i.e. an input inside a modal with z-index: 1000), the element does get highlited properly, but interaction in the element is not posible, even if disableInteraction: false is explicitly set in the option's configuration. 

## Expected Behavior
Interaction should be allowed on highlited elements inside another stacking context.

## Actual Behavior
Interaction does not work in this case, clicking the highlited element does nothing.

## Errors and Screenshots 
When right click -> inspecting the highlited box, dev tools takes me to the div with "introjs-helperLayer" class, oposed to normal behaviour in which inspecting the box takes me to the actual element. For what I have seen, this happens because, even when the element gets the "showElement" class that gives it an enormous z-index, it is still in the modal's stacking context, so it cannot be put over the helperLayer which exists in the root stacking context. I've read through the docs to see if there's an existing solution to this problem, but I haven't found anything (don't know if there's actually a common solution that I have overlooked).
![image](https://user-images.githubusercontent.com/35577671/113769617-6bc65c00-96de-11eb-949e-e8dc1eee68c6.png)
![image](https://user-images.githubusercontent.com/35577671/113769820-a3350880-96de-11eb-9695-e2f74a14eb25.png)

## Environment (optional)
Tried on both Chrome and Firefox, introjs version is 3.4.0, we're developing an Angular application using PrimeNG.

ajjones20:
I am having this same issue. I want interactivity in a bootstrap modal, rearranging the DOM doesn't seem to resolve the problem either.

afshinm:
I can look into this but I believe this is an exception where element interaction doesn't work because you have an absolute/fixed parent. 

Let me know if anyone has any suggestions.

sdraper69:
I too get this issue, I agree it seems to be because the target's parent is absolutely positioned.

Unfortunately this makes intro unusable for me. It sounds like there is no workaround? 

If in chrome dev tools I take untick the position: absolute then I can see its actually selectable, to me this would suggest that however you are creating the "hole" to click it just isn't looking at the elements actual position.

KevinMoran99:
> I can look into this but I believe this is an exception where element interaction doesn't work because you have an absolute/fixed parent.
> 
> Let me know if anyone has any suggestions.

[...]

PR #1258, opened 2021-05-21
chore: adding tests to capture the stacking contexts issue

afshinm:


afshinm:
https://github.com/usablica/intro.js/issues/1202

PR #1266, opened 2021-05-24
Allows for local step asynchronous events

ezra-obiwale:
- Individual steps can now provide `onbeforechange(next)` and `onchange()` event callbacks
- When provided, the local events would be called while the global events would not.

Bonus:

- Adds class `no-title` to the tooltip header layer if a step doesn't have a title
- Adds step class (step-[STEP_NUMBER]) to the reference layer

### DIFF
--- a/src/core/introForElement.js
+++ b/src/core/introForElement.js
@@ -135,7 +135,7 @@ export default function introForElement (targetElm, group) {
             element: currentElement,
             title: currentElement.getAttribute("data-title") || "",
             intro: currentElement.getAttribute("data-intro"),
-            step: parseInt(currentElement.getAttribute("data-step"), 10),
+            step,
             tooltipClass: currentElement.getAttribute("data-tooltipclass"),
             highlightClass: currentElement.getAttribute("data-highlightclass"),
             position:
--- a/src/core/introForElement.js
+++ b/src/core/introForElement.js
@@ -16,8 +16,7 @@ import createElement from "../util/createElement";
  * @param {String} group
  * @returns {Boolean} Success or not?
  */
-export default function introForElement(targetElm, group) {
-  const allIntroSteps = targetElm.querySelectorAll("*[data-intro]");
+export default function introForElement (targetElm, group) {
   let introItems = [];
 
   if (this._options.steps) {
@@ -68,6 +67,9 @@ export default function introForElement(targetElm, group) {
       }
     });
   } else {
+    const allIntroSteps = targetElm.querySelectorAll("*[data-intro]");
+    const introWithStepCount = allIntroSteps.filter(elem => elem.getAttribute('data-step') !== null).length;
+
     //use steps from data-* annotations
     const elmsLength = allIntroSteps.length;
     let disableInteraction;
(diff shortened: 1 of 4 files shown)


src/core/showElement.js
```diff
@@ -94,6 +95,22 @@ export default function _showElement(targetElement) {
...
+    oldReferenceLayer.classList.remove(`step-${targetElement.step - 1}`);
+    // remove next step class in case going back
+    oldReferenceLayer.classList.remove(`step-${targetElement.step + 1}`);
+    // add current step class
+    oldReferenceLayer.classList.add(`step-${targetElement.step}`);
+
+    if (targetElement.title) {
+      oldTooltipHeaderLayer.classList.remove('no-title');
+    } else {
+      oldTooltipHeaderLayer.classList.add('no-title');
```
afshinm, line 112: How this `no-title` class will be used? I couldn't find any CSS changes in this PR.

src/core/introForElement.js
```diff
@@ -146,35 +130,31 @@ export default function introForElement(targetElm, group) {
...
-          highlightClass: currentElement.getAttribute("data-highlightclass"),
-          position:
-            currentElement.getAttribute("data-position") ||
-            this._options.tooltipPosition,
-          scrollTo:
-            currentElement.getAttribute("data-scrollto") ||
-            this._options.scrollTo,
-          disableInteraction,
-        };
+        if (step > 0) {
```
afshinm, line 133: This is a scary function to change mainly because we don't have any tests for it.

src/core/showElement.js
ezra-obiwale: Yeah, it's for themes to use if they need it.

For instance, in my use case, I needed to hide the header element if there's no title.

src/core/introForElement.js
ezra-obiwale: Yeah, the point is to use a single loop, instead of two. As such, existing tests already cover this.

afshinm: Not sure if we have tests for this? there are two ways of adding the steps, using the data-* attributes and a JSON object. I don't think we have any tests for the data-* attr approach.

src/core/showElement.js
afshinm: Right, got it. I personally think we should keep the codebase simple and avoid adding any classes that is not used right now. I understand that you need that no-title class in _your_ application though. Can you use your initial input (e.g. your steps list) to see if a specific step as title or not?

src/core/introForElement.js
afshinm: I just started refactoring this function and adding some tests. I will ping you once it's merged.

afshinm: WIP https://github.com/usablica/intro.js/pull/1295

src/core/showElement.js
ezra-obiwale: Yeah, thought of that but that's JS. It's better to use CSS if possible instead of JS.

About avoiding adding classes, it's not unusual for libraries to add css classes that may help developers customize the outlook, even if the library doesn't use the classes.

DamienCassou:
I would be interested in this feature.

PR #1307, opened 2021-06-16
feat: updating the bullets layer after refresh(true)

afshinm:


AILazerka:
Nice. Thank you a lot. Great job 👍

Issue #861, opened 2018-05-17
Resizing the page throws errors


plentylife:
## Description
When resizing the page, there are TypeErrors that are thrown

## Expected Behavior
No errors occur

## Actual Behavior
This error is thrown
`Uncaught TypeError: Cannot read property 'style' of null
    at IntroJs._placeTooltip (index.js:14449)
    at IntroJs._refresh (index.js:14339)
    at IntroJs.refresh (index.js:16259)
    at IntroJs._onResize (index.js:14128)
    at handler (index.js:15422)`

saumilcmarix:
Hello,
I got a strange error if I click on tutorial multiple time, which says
Uncaught TypeError: Cannot read property 'setAttribute' of null
    at IntroJs._showElement (intro.js:1367)
    at IntroJs._nextStep (intro.js:444)
    at IntroJs._introForElement (intro.js:270)
    at IntroJs.start (intro.js:2370).

Please refer screenshot: [http://prntscr.com/nz73s6](url)

gp0:
I know this super-old, but i have just run into this. The cause seems to be that _onResize blindly tries to refresh, and if the .introjs-tooltip does not exist, it fails. I've worked around this with: 
`function _onResize () {
    if (document.querySelector('.introjs-tooltip') !== null) {
      this.refresh.call(this);
    }
  }`

robsoncombr:
Hello, I would like to register that I have also just run into this. On my case it happens inside _refresh():

Uncaught TypeError: Cannot read property 'style' of null
    at IntroJs._placeTooltip (intro.js?479f:603)
    at IntroJs._refresh (intro.js?479f:493)
    at IntroJs.refresh (intro.js?479f:2413)
    at IntroJs._onResize (intro.js?479f:282)
    at handler (intro.js?479f:1576)
    (..)

When it calls the following method with null values since the tooltip does not exist:
      _placeTooltip.call(this, this._introItems[this._currentStep].element, oldtooltipContainer, oldArrowLayer, oldHelperNumberLayer);

Can we please have this fixed in methods _onResize and _refresh?

Thank you veru much.

jmeinke:
I can confirm this error.

Issue #1282, opened 2021-06-02
tooltipLayer undefiend when resize window size after call `Introjs().exit()` 


0721Betty:
## Description
resize window size after call `Introjs().exit()`

## Expected Behavior
no issue

## Actual Behavior
![image](https://user-images.githubusercontent.com/48433973/120463982-02f21c00-c3cf-11eb-95f9-c404999557fa.png)

## Errors and Screenshots (optional)
![image](https://user-images.githubusercontent.com/48433973/120464461-8875cc00-c3cf-11eb-8faf-847d9394354d.png)

![image](https://user-images.githubusercontent.com/48433973/120464331-62502c00-c3cf-11eb-8503-fbdeb6d61871.png)

![image](https://user-images.githubusercontent.com/48433973/120464557-a6433100-c3cf-11eb-80a6-fac849a468e0.png)

## Environment (optional)
chorme, introjs: ^2.9.3, mac system, vue

0721Betty:
If I add the code circled in the screenshots，it will not report an error
![image](https://user-images.githubusercontent.com/48433973/120594129-0ccd5b00-c473-11eb-8921-b6475b765430.png)

0721Betty:
The latest version ^3.4.0 has the same error

afshinm:
I was not able to reproduce this. I ran introJs.start(), then introJs.exit() and resized the window but didn't get the same error.

PR #1330, opened 2021-07-06
(Fix) On resize method throws error when IntroJS is not visible on the page

nicobarengo:
This fixes Issue #861

### DIFF
--- a/src/core/onResize.js
+++ b/src/core/onResize.js
@@ -1,5 +1,7 @@
 import refresh from "./refresh";
 
 export default function onResize() {
-  refresh.call(this);
+  if (document.querySelector('.introjs-tooltip') !== null) {
+    refresh.call(this);
+  }
 }


src/core/onResize.js
```diff
@@ -1,5 +1,7 @@
 import refresh from "./refresh";
 
 export default function onResize() {
-  refresh.call(this);
+  if (document.querySelector('.introjs-tooltip') !== null) {
```
afshinm, line 4: I think it makes more sense to move this if-else to the `refresh` function.

afshinm (changes requested):

src/core/onResize.js
walkermalling: I'm pretty sure this would fix https://github.com/usablica/intro.js/issues/1282 (as well as https://github.com/HiDeoo/intro.js-react/issues/70)

walkermalling: > I'm pretty sure this would fix #1282 (as well as [HiDeoo/intro.js-react#70](https://github.com/HiDeoo/intro.js-react/issues/70))

After some more digging, I believe these are separate error paths. The bug in the ticket I referenced is triggered when there is no element on the current step (https://github.com/usablica/intro.js/blob/master/src/core/refresh.js#L43), whereas this PR would correct for a missing `.introjs-tooltip`, which is the next parameter in `placeTooltip` (https://github.com/usablica/intro.js/blob/master/src/core/refresh.js#L44).

afshinm:
Fixed in this commit https://github.com/usablica/intro.js/commit/5c43bc9f21cc8490f6b9f4764bbe694bca2c471b

PR #1438, opened 2021-09-29
prevent errors on refresh

walkermalling:
This responds to the suggestion in @nicobarengo 's PR (https://github.com/usablica/intro.js/pull/1330), and covers additional error cases. Two cases have documented issues: when no element matches `.introjs-tooltip` and when the `element` property is null. Both of these issues cause `placeTooltip` to throw. 

Related: 
- https://github.com/usablica/intro.js/issues/861
- https://github.com/HiDeoo/intro.js-react/issues/70
- https://github.com/usablica/intro.js/issues/1282

### DIFF
--- a/src/core/refresh.js
+++ b/src/core/refresh.js
@@ -38,6 +38,18 @@ export default function refresh(refreshSteps) {
     const oldArrowLayer = document.querySelector(".introjs-arrow");
     const oldtooltipContainer = document.querySelector(".introjs-tooltip");
 
+    if (!this._introItems[this._currentStep].element) {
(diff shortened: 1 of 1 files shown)


afshinm:
Thanks @walkermalling! Could you add at least one test for this fix please?

afshinm:
https://github.com/usablica/intro.js/commit/5c43bc9f21cc8490f6b9f4764bbe694bca2c471b

We already check the currentStep here https://github.com/usablica/intro.js/blob/master/src/core/placeTooltip.js#L196

Issue #1485, opened 2021-11-08
Translate "of" in helperNumberLayer


dmitry53:
How its possible to translate or change "of" ?
![Снимок экрана 2021-11-09 023224](https://user-images.githubusercontent.com/2632988/140834838-dfee7c86-deed-4060-ba88-1ae8c5b48f33.png)

dmitry53:
any idea? I need use 1 / 18 step

ivan-jumbomail:
> any idea? I need use 1 / 18 step

1. Add handler for onBeforeChange
2.  Find element represents it, for example `const stepNumsEl = document.querySelector('.introjs-helperNumberLayer');`
3.  Set data-attr to the elment like `stepNumsEl.setAttribute('data-step-nums', `${nextStepIndex + 1} / ${steps.length}`);`
4.  Add styles  like `
 .introjs-helperNumberLayer:before {
  display: block;
  position: absolute;
  bottom: 14px;
  content: attr(data-step-nums);
}
`
5. Hide previous text.
![image](https://user-images.githubusercontent.com/97881234/151795323-a27cfcc3-ca9e-4889-98cd-7c8a97179fd6.png)

dmitry53:
> 4\. .introjs-helperNumberLayer:before {   display: block;   position: absolute;   bottom: 14px;   content: attr(data-step-nums); }

thank you Ivan, it works!

`     
      const intro = introJs();
      intro.setOptions({
        tooltipClass: 'introjs-vue',
        showBullets: false,
        showProgress: true,
        showStepNumbers: true,
        steps: this.IntroSteps,
      });

intro.onafterchange(() => {
        const stepNumsEl = document.querySelector('.introjs-helperNumberLayer');
        stepNumsEl.setAttribute('data-step-nums', `${intro._currentStep + 1} / ${intro._introItems.length}`);
      }).start();
`
css
`
.introjs-helperNumberLayer:before {
  display: block;
  position: absolute;
  bottom: -2px;
  content: attr(data-step-nums);
  font-size: 11px;
  white-space: nowrap;
}
`

Issue #1326, opened 2021-06-29
Add option to localize showStepNumbers label


hmdhasani:
In case of enabling showStepNumbers the displying label is always in format of "n of m" and there is no way to localize it for other languages. I think below code need to be localizedable:

https://github.com/usablica/intro.js/blob/053d17b40bca4b74a7d24547970484038915a64a/src/core/showElement.js#L365

afshinm:
thanks @hmdhasani. Let me know if you want to work on this. This should be an easy one to fix.

hmdhasani:
I'm afraid I have no time for that now. I just changed it in mine file. I will tell here if in future I have time for that.
Another thing that needs to be localized is usage of arrow keys as shortcuts of previous/next. I thinks it's better if it can be swapped for RTL languages.

PR #1588, opened 2022-02-16
custom step number separator

aperezf:
Added new option: customStepNumbersSeparator. Use custom separator when showStepNumbers is true.

Related:
#1485 
#1326


afshinm:
This is now available https://github.com/usablica/intro.js/blob/master/src/index.js#L67

PR #1597, opened 2022-02-25
Add next/previous callbacks

### LINKED ISSUE #1533: Add nextButton callback when user clicked
## Description
I want to add callback when user clicked nextButton，but i found there is no interface to set，can u guys consider expose it

# Expected Behavior
What did you expect to happen?

## Actual Behavior
What actually happened?

## Errors and Screenshots (optional)
Give us any additional information if you have.

## Example (recommended)
How can someone else see/test this issue?  List the steps or write some code; or create a jsfiddle or codepen.

## Environment (optional)
List browser information, Intro.js version, Operating System, JavaScript libraries (jQuery, Angular, React), etc.

neurot1k:
### What?
I've added callbacks for next and previous, including tests and an example.

### Why?
I wanted a callback that was run before the step was changed and indicated the direction of change.

onbeforechange is called after the step is changed but before it is shown (this.currentStep() returns the new step in onbeforechange) and does not tell you the direction.

Closes #1533

### DIFF
--- a/example/callbacks/onnext-onprev.html
+++ b/example/callbacks/onnext-onprev.html
@@ -73,11 +73,11 @@
         var intro = introJs();
         intro.onnext(function () {
           const nextStep = this.currentStep() + 1;
-          alert(`Next up: ${this._introItems[nextStep].intro}`);
+          return confirm(`Next intro: ${this._introItems[nextStep].intro}\nProceed?`);
         });
         intro.onprevious(function () {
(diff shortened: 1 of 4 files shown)


neurot1k:
I've decided not to use/support this library due to the licensing and the number of issues I've run into during my evaluation. As a result I've closed this PR.

neurot1k:
I hope you find this useful and merge it, I just no longer wish to be involved. Thanks!

PR #1608, opened 2022-03-15
added hideSkip option

yyc-github:
Added hideSkip option

### DIFF
--- a/src/core/showElement.js
+++ b/src/core/showElement.js
@@ -458,6 +458,16 @@ export default function _showElement(targetElement) {
     _disableInteraction.call(self);
   }
 
+  // hide skip button
+  if (this._options.hideSkip === true) {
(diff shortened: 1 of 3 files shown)


src/core/showElement.js
```diff
@@ -458,6 +458,16 @@ export default function _showElement(targetElement) {
     _disableInteraction.call(self);
   }
 
+  // hide skip button
+  if (this._options.hideSkip === true) {
```
afshinm, line 462: Could you add a test case for this if-statement please?

afshinm (changes requested):

src/core/showElement.js
yyc-github: @afshinm added two test cases, please have a look.

PR #1679, opened 2022-05-26
Query element by selector later if it's not exists during configuration

### LINKED ISSUE #1674: Feature request: query element by selector on step if it's not exists during configuration
Would  it be handy to get element by `document.querySelector` just before or at the begining of his step if it doesn't exists during configuration?

I have such situation and want to implement this feature, cause `introJs.refresh(true)`:

- Could be non-optimal to fetch all elements of steps several times
- It's not handy to call `introJs.refresh(true)` on each pre-step
- `introJs.refresh(true)` don't position properly (I haven't figured out why yet)

gooddaytoday:
I'm sorry I didn't wait for approval #1674 :racing_car: . Just need to buy commercial license with this feature :slightly_smiling_face: .  

Processing possible absence of an element, configured with string selector: when creating list of steps, we take into account the possible absence of element.

- If `step.element` is a selector for which there is no element at the configuration stage, then we replace `.element` specifying a getter/setter for it. In the getter  we return the element via `document.querySelector`
- Since that moment we assume that `.element` should return `HTMLElement` by selector and if there is still no element - throwing `Error`. Thus to check raw field below we need use `._element`. 

The behavior has changed: earlier if there wasn't element by selector, IntroJS silently created floating tooltip at that step. 

Also:

- [Fixed one](/gooddaytoday/intro.js/commit/7112475364e16cdb055ad58da8408d752a9fea2c) test cause of changed behaviour
- Added some [unit tests](/gooddaytoday/intro.js/commit/fc3cdfde766b583959fe4dc2ee1fd110a9d1fe4f)
- And Cypress: [button to create div](/gooddaytoday/intro.js/commit/9031e45c39285dff90c6518ae8250f6d3cb4e5ac), [test](/gooddaytoday/intro.js/commit/0a60954c56ad787eefb37abb33243a4c3b10cb6a) and [updated snapshots](/gooddaytoday/intro.js/commit/7fc38b63a138263a9f0632b23b993ea98efaa882)

Closes #1674.


gooddaytoday:
And it's not finished yet. If element is added with delay (for example, after async request to server), an error will occur

gooddaytoday:
> And it's not finished yet. If element is added with delay (for example, after async request to server), an error will occur

Finished that feature:

- Added function to wait until element appears in DOM https://github.com/usablica/intro.js/pull/1679/commits/79dea703cc1a605ba956cebbfb4cda828dac434f
- `waitforElement` function applied in `nextStep` https://github.com/usablica/intro.js/pull/1679/commits/b6e885a579cfdc809b6f92159b4f814fe70d72fd

gooddaytoday:
> * `waitforElement` function applied in `nextStep` [b6e885a](https://github.com/usablica/intro.js/commit/b6e885a579cfdc809b6f92159b4f814fe70d72fd)

Fixed commit https://github.com/usablica/intro.js/pull/1679/commits/b61ffab232720e0ada79c204ed84fd05467b5e78  and added test https://github.com/gooddaytoday/intro.js/commit/8716a1a038324fad58af5f9ba3d7cd9da6f65ac5  for that error

tests/core/fetchIntroSteps.test.js
```diff
@@ -80,11 +81,44 @@ describe("fetchIntroSteps", () => {
...
+    );
+
+    expect(steps.length).toBe(1);
+
+    try {
+      const element = steps[0].element; // jshint ignore:line
+    } catch (e) {
+      if (!e.message.includes("There is no element with given selector:"))
+        throw e;
+    }
```
afshinm, line 113: Are we checking here to make sure the steps array is empty?

src/core/fetchIntroSteps.js
```diff
@@ -26,14 +26,37 @@ export default function fetchIntroSteps(targetElm) {
...
-
-      //intro without element
-      if (
+        const el = document.querySelector(currentItem.element);
+        if (el !== null) {
+          currentItem.element = el;
+        } else {
+          // If element is not exists yet, we'll get it on step
+          const elSelector = currentItem.element;
+          Object.defineProperty(currentItem, "element", {
```
afshinm, line 35: Is it important to have this object when the DOMElement doesn't exist? Can we just replace it with `undefined`?

src/util/waitForElement.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param {() => void} callback Callback to be called after Element appearance
+ */
+export default function _waitForElement(elSelector, callback) {
+  if (document.querySelector(elSelector) !== null) {
+    callback();
+    return;
+  }
+
+  if (typeof MutationObserver !== "undefined") {
+    /* jshint ignore:start */
```
afshinm, line 16: Why are we suppressing jshint errors? Happy to help if there are any issues with our jshint setup

src/util/waitForElement.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+/**
+ * @param {string} elSelector
+ * @param {() => void} callback
+ * @param {number} checkInterval In milliseconds
+ * @param {number} maxTimeout In milliseconds
+ */
+function waitForElementByTimeout(
```
afshinm, line 43: Can we test this function please? I'm a little concerned since it's doing a recursive call.

tests/cypress/integration/tour/navigation.js
```diff
@@ -47,6 +47,14 @@ context("Navigation", () => {
     cy.get(".introjs-showElement").should("not.exist");
   });
 
+  it("should exit the tour after right btn pressed at the end", () => {
+    cy.get(".introjs-tooltiptext").contains("step one");
```
afshinm, line 51: Beautiful! Thank you

tests/core/fetchIntroSteps.test.js
gooddaytoday: No, we calling here `.element` of the first array item to check that Error "There is not element..." is throwing in case of element absence.

src/core/fetchIntroSteps.js
gooddaytoday: Yes, my first (local) implementation was with `undefined`, but it could be more changes in other code, that calls `.element`. 

I'll try to do with `undefined`.

src/util/waitForElement.js
gooddaytoday: Yep, it could be deleted. Added MutationObserver to globals in jshint config and removed jshint ignore (33e560b3f576d99f9d134cc6a317e9945c6f0082)

tests/cypress/integration/tour/navigation.js
gooddaytoday: In fact, there wasn't such error in master before adding this feature.  Possible error was  taken into account in 87fa6043b8c08d3d19d397160896ea15bd5fb467

src/util/waitForElement.js
gooddaytoday: Added tests in https://github.com/gooddaytoday/intro.js/commit/d5eaf0d3d2f29f3fc6d934c1e99190954b48181a

skuracj:
@gooddaytoday @afshinm  
Is this feature ready for merge?

Issue #1703, opened 2022-06-29
Close button misplaced when using long title


sammuell:
## Description
When defining a long title which wraps onto multiple lines, the close button does not stay in the corner. Instead, it is shown below the title.
It seems to be a floating issue. A solution might be to use a flexbox instead of floating the title and close button.

## Expected Behavior
The close button is expected to always stay in the top right corner.

## Actual Behavior
![grafik](https://user-images.githubusercontent.com/1202195/176441021-153b468a-3838-420c-b980-5247304db1dc.png)
![grafik](https://user-images.githubusercontent.com/1202195/176441141-9b4f3c76-b957-4da8-bddb-c48888f1866c.png)

## Example (recommended)
Use the following code on this page: https://introjs.com/docs/examples/basic/json-config
```js
introJs().setOptions({
  steps: [{
    title: 'Welcome Welcome Welcome Welcome Welcome Welcome Welcome',
    intro: 'Hello World! 👋'
  },
  {
    element: document.querySelector('.card-demo'),
    title: 'card-demo card-demo card-demo card-demo card-demo card-demo card-demo card-demo card-demo',
    intro: 'This step focuses on an image'
  },
  {
    title: 'Farewell Farewell Farewell Farewell Farewell Farewell Farewell Farewell!',
    element: document.querySelector('.card__image'),
    intro: 'And this is our final step!'
  }]
}).start();
```

PR #1708, opened 2022-07-05
Correct position of skip button when having multiline titles

sammuell:
Addressing the issue #1703


afshinm:
This is great. Thanks @sammuell.

Issue #1719, opened 2022-07-14
UX: page should scroll smoothly


sammuell:
## Description
The element indicator uses an animation (css transistion) to move to the next element. However, if the element is outside of the viewport the page scrolls instantly to the element. It is confusing because the animated element leaves the viewport which then jumps to another location.

## Expected Behavior
It would be nice if both would be smoothly animated or none

## Actual Behavior
The page scroll is instant.

## Potential solution
Use [scroll-behavior](https://developer.mozilla.org/de/docs/Web/CSS/scroll-behavior) on html tag (and scrollable container elements) using a separate css class which is applied temporarily. Use the same class to apply the transition to the helper element. Add an option to apply this class, making both animations optional.

Zvonimir1995:
Recently, i needed to implement a smooth scroll in my introJS tutorial, and, as a workaround, you can use introJS onbeforechange function. It looks something like this:

 ```
introJs().onbeforechange(async (htmlEl) => {
 	await scrollToElement(htmlEl, mdDown);
         return true;
 }).setOptions({...}).start()
```
 

Then, on your element you can data-custom-scroll-to property and set it to something meaningful to you. For instance, I use "top" or "element", for "top", i scroll to top: 0, and for "element", I scroll to top: htmlEl.offsetTop - 300 .

My function looks like this:

```
 export const scrollToElement: (htmlEl: HTMLElement, mdDown: boolean) => Promise<true> = (
 	htmlEl,
 	mdDown
 ) => {
 	return new Promise((resolve) => {
 		const customScrollAttribute = htmlEl.getAttribute(
 			'data-custom-scroll-to'
 		) as CustomScrollTo | null;
 		if (customScrollAttribute === 'top') {
 			window.scrollTo({
 				top: 0,
 				left: 0,
 				behavior: 'smooth'
 			});
 			delay(250).then(() => {
 				resolve(true);
 			});
 		} else if (customScrollAttribute === 'element') {
 			window.scrollTo({
 				top: htmlEl.offsetTop - (mdDown ? 250 : 300),
 				left: 0,
 				behavior: 'smooth'
 			});
 			delay(250).then(() => {
 				resolve(true);
 			});
 		} else {
 			resolve(true);
 		}
 	});
 };
```

In this case, my htmlEl look like:
`<section data-custom-scroll-to={'element' as CustomScrollTo} className="...">`

Other solution, you can add smooth scrolling globally if you want. But I did not want to do that.
[...]

PR #1728, opened 2022-07-18
Disable tour animation

sammuell:
This PR adds an option to disable any transitions of the tour (on by default). It also implements smooth page scrolling (issue #1719)

### DIFF
--- a/src/util/scrollTo.js
+++ b/src/util/scrollTo.js
@@ -30,21 +30,25 @@ export default function scrollTo(scrollTo, { element }, tooltipLayer) {
     // the center of the target element or tooltip.
 
     if (top < 0 || element.clientHeight > winHeight) {
-      window.scrollBy(
-        0,
-        rect.top -
+      window.scrollBy({
+        top: rect.top -
           (winHeight / 2 - rect.height / 2) -
-          this._options.scrollPadding
-      ); // 30px padding from edge to look nice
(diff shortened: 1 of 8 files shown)


src/util/scrollTo.js
```diff
@@ -30,21 +30,25 @@ export default function scrollTo(scrollTo, { element }, tooltipLayer) {
     // the center of the target element or tooltip.
 
     if (top < 0 || element.clientHeight > winHeight) {
-      window.scrollBy(
-        0,
-        rect.top -
+      window.scrollBy({
```
afshinm, line 33: Would be great if we add some Cypress tests

afshinm (commented):
Great work! The only missing piece seems to be the tests which should be fairly straightforward to add. Happy to help

src/util/scrollTo.js
sammuell: Agreed. What would you test in this case? It still uses the same function window.scrollBy()

afshinm: Sorry for the delay @sammuell. We usually have a delay() between each step in the Cypress tests to make sure the transition is complete. Now that the transition can be enabled/disabled, we can probably have a test _without_ that delay(), take a screenshot and that should capture the tooltip in its correct position. Would that be a useful test case?

PR #1753, opened 2022-08-17
A11y improvements

### LINKED ISSUE #1752: Improve a11y by using button instead of links
## Description
Links should only be used when the URL change.
`<a role="button">` is not the same as `<button type="button">`.

## Expected Behavior
If you are creating a button, use a `<button>`.

## Actual Behavior
The library generate links with no `href` and `role="button"`

mrleblanc101:
Fix #1752


mrleblanc101:
I should probably have opened this as Draft while I work on fixing the test.

mrleblanc101:
I have 9 out of 24 test failing even on master on my machine saying the snapshot are too different.
Is the 0.05 threshold too low ?

mrleblanc101:
I have only 6 test failing, all with less than 0.07 variance on Github Action.
On my local machine, I'm not able to see that actual variance to fix it.
Do you think it would be possible to update the snapshot @afshinm ?

mrleblanc101:
@afshinm Just let me know if you see anything visual regression I could fix, otherwise this PR is ready.

afshinm:
Sounds good, thanks @mrleblanc101. Will have a look.

lilsweetcaligula:
@afshinm @mrleblanc101 Hey guys. I just wanted to point out, - this PR also fixes a related bug, wherein the anchor/button element had its `role="tab"` overriden with `role="button"`.

https://github.com/usablica/intro.js/blob/v6.0.0/src/core/showElement.js#L79
https://github.com/usablica/intro.js/blob/v6.0.0/src/core/showElement.js#L87
https://github.com/usablica/intro.js/blob/v6.0.0/src/util/setAnchorAsButton.js#L8

The bug is a problem because it poses a critical Web Accessibility error.

Issue #951, opened 2019-11-12
TypeError when calling _nextStep function on last introItem by pressing right arrow key.


tonykiselis:
## Description
There is a bug in `_nextStep` method which results in:
```
Cannot read property 'element' of undefined
```

## Expected Behavior
`_nextStep` should not execute when right arrow is pressed while on last item in introItem set.

## Actual Behavior
When one uses arrow keys to move through items, after reaching last element on the item set and pressing right arrow key produces an error, because this._currentStep exceeds the this._introItems array, resulting in call on `undefined`.

## Errors and Screenshots (optional)
![error_intro_js](https://user-images.githubusercontent.com/27768115/68668782-a3fd5580-0551-11ea-8b34-3ae94b4a89a5.png)

## Example (recommended)
Press right arrow key while being on last item in the set of Intro js

## Environment (optional)
Problem is with code logic, not the environment.

Problem can be solved by simple if check:

```
    if (typeof (this._currentStepNumber) !== 'undefined') {
      _forEach(this._introItems, function (item, i) {
        if( item.step === this._currentStepNumber ) {
          this._currentStep = i - 1;
          this._currentStepNumber = undefined;
        }
      }.bind(this));
    }

    // added check here before _currentStep mutation
    if (this._introItems.length - 1 < this._currentStep + 1) {
      return;
    }

    if (typeof (this._currentStep) === 'undefined') {
      this._currentStep = 0;
    } else {
      ++this._currentStep;
    }
```

Adifmac:
This does not throw an error, but ignores the user expectation, it should close the intro.
In case you want to close intro - simply add exit command:
`// added check here before _currentStep mutation
	if (this._introItems.length - 1 < this._currentStep + 1) {
		this.exit();
		return;
}`

PR #1885, opened 2023-03-09
Fix undefined access of nextStep.element

reuben:
See https://github.com/usablica/intro.js/issues/951
See https://github.com/HiDeoo/intro.js-react/issues/33


afshinm:
@reuben thanks. could you rebase your branch please?

PR #1934, opened 2023-06-13
Fix getOffset function to render elements in iframe offset from the p…

Coder2002-MSAdithya:
This fix does NOT affect any other parts of the project and is also backward compatible. This idea was originally given by arjshiv(Arjun Kannan) back in 2015

### DIFF
--- a/src/util/getOffset.ts
+++ b/src/util/getOffset.ts
@@ -28,6 +28,16 @@ export default function getOffset(
     height: x.height,
   };
 
+  let boundingFrameClientRect = {
+    top: 0,
+    left: 0
+  }
+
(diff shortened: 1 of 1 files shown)


src/util/getOffset.ts
```diff
@@ -28,6 +28,16 @@ export default function getOffset(
     height: x.height,
   };
 
+  let boundingFrameClientRect = {
+    top: 0,
+    left: 0
+  }
+
+  if((element) && (element.ownerDocument !== window.document))
```
afshinm, line 36: Could you add e2e tests for this?

afshinm (changes requested):

Mouerr (approved):

PR #1950, opened 2023-07-06
fix: behavior of unassigned 'intro' property in json

oferze:
Currently, one has to set the `intro` property in the steps json. Changed it to behave like `title`, that is, it's not compulsory. A step could have a title only.

Without this change, a step would contain "undefined":

image


tinyappz:
This is now fixed in `master`. Thanks @oferze

PR #1993, opened 2023-08-17
Check type of jQuery var

miskith:
Check the type of jQuery variable before calling instanceof in case for some reason jQuery is defined but is not function


binrysearch:
This is fixed now

PR #2029, opened 2023-12-02
fix: `cloneObject` when `window.jQuery` set but null/undefined

atoff:
I encountered a problem when testing locally with `intro.js` whereby (for some reason) `window.jQuery` was set on my browser to `undefined`. This causes an exception in cloneObject as it tries to use the value for an instance of comparison.
image

My app doesn't use jQuery, so not sure where it is coming from, but either way I think we need to assert that window.jQuery is non-nullable before continuing to the `instanceof`.

I only encountered this problem when defining the `steps` option, as `cloneObject` is used to clone the input.

### DIFF
--- a/src/util/cloneObject.ts
+++ b/src/util/cloneObject.ts
@@ -11,7 +11,7 @@ export default function cloneObject<T>(source: T): T {
 
   for (const key in source) {
     // @ts-ignore:next-line
-    if ("jQuery" in window && source[key] instanceof window.jQuery) {
+    if ("jQuery" in window && window.jQuery && source[key] instanceof window.jQuery) {
       temp[key] = source[key];
     } else {
       temp[key] = cloneObject(source[key]);


jeroensijbom (approved):

kingsloi:
@jeroensijbom I see you approved this PR, but is there any update on when it'll be merged? I'm not able to use the intro.js without first including jQuery, which the project I'm using intro.js in, doesn't use, so we're loading jQuery to not really use it

kingsloi:
cheers @atoff! I was running into the same issue with a the TALL stack.

@jeroensijbom I see you approved this PR recently, but is there any update on when it'll be merged? I'm not able to use the intro.js without first including jQuery, which the project I'm using intro.js in doesn't use, so we're loading jQuery to not really use it, and removing it breaks it 🙏

atoff:
> cheers @atoff! I was running into the same issue using intro.js with a project made in the TALL stack.
> 
> @jeroensijbom I see you approved this PR recently, but is there any update on when it'll be merged? I'm not able to use the intro.js without first including jQuery, which the project I'm using intro.js in doesn't use, so we're loading jQuery to not really use it, and removing it breaks it 🙏

Workaround I am using at the moment is including
```js
  delete window.jQuery
```

when I bootstrap my application - should work for you until this is merged and released!

kingsloi:
> Workaround I am using at the moment is including
> 
> ```js
>   delete window.jQuery
> ```
> 
> when I bootstrap my application / use intro.js in a component - should work for you until this is merged and released!

Amazing @atoff! Worked a treat - cheers mind kind sir! 🍻

binrysearch (approved):

PR #2064, opened 2024-07-13
Restructure the project and public APIs

binrysearch:
The goal of this PR is to make the project modular and easier to maintain and update.

- Refactors the project to be modular, with each part of the library added to `./packages` (e.g. `./packages/tour`)
- Updates Cypress and Jest tests to be colocated within each package
- Adds new Cypress commands to run the tests locally in a docker container


afshinm (approved):
lgtm! We should refactor the tooltip package as well.

PR #2091, opened 2025-03-10
fix: properly render HTML content in tooltips using safe template parsing

Parvinmh:


src/packages/tour/components/TourTooltip.ts
```diff
@@ -451,8 +451,13 @@ export const TourTooltip = ({
...
   children.push(Header({ title, skipLabel, onSkipClick }));
 
-  children.push(div({ className: tooltipTextClassName }, text));
-
+  const template = document.createElement("template");
+  template.innerHTML = text;
+  
+  const parsedContent = document.createRange().createContextualFragment(text);
+  
+  children.push(div({ className: tooltipTextClassName }, parsedContent));
```
binrysearch, line 459: This should be moved to the `Tooltip` component instead of `TourTooltip`. Reason being, the `Tooltip` component will be used by other consumers (e.g. Hint) which must support HTML rendering.

PR #2102, opened 2025-04-21
Fix/deprecated properties

Parvinmh:


src/packages/hint/components/HintTooltip.ts
```diff
@@ -15,6 +16,7 @@ export type HintTooltipProps = Omit<
   closeButtonLabel: string;
   closeButtonClassName: string;
   className?: string;
+  renderAsHtml?: boolean;
```
binrysearch, line 19: Is this supposed to be part of this change?

binrysearch (changes requested):

Issue #2104, opened 2025-06-23
Need of Exit Button to exit entire Tour


PatilHarshh:
There is is need of additional button. " Exit"
condition : Our Org started using introjs. and I am working on this Part. we had long intros. there are need of Exit button for this.

Assign it to me i will work.
email: harshpatil2956@gmail.com

@Parvinmh

PatilHarshh:
![Image](https://github.com/user-attachments/assets/4f5ff59d-260b-4dce-944d-0e17d8bcf4b0)

this would be fine @Parvinmh ? for your UI?

Parvinmh:
@PatilHarshh Yes, this works fine for now. I can later work on improving the UI if needed.

PatilHarshh:
@Parvinmh please review

binrysearch:
Replied in the PR. We should work on providing a way to customise the dialog in future versions. This should be doable but requires a bit of work

PR #2125, opened 2025-07-20
added exit button in the tour

PatilHarshh:
image

### DIFF
--- a/src/packages/tour/components/TourTooltip.ts
+++ b/src/packages/tour/components/TourTooltip.ts
@@ -10,6 +10,7 @@ import {
   helperNumberLayerClassName,
   hiddenButtonClassName,
   nextButtonClassName,
+  exitButtonClassName,
   previousButtonClassName,
   progressBarClassName,
   progressClassName,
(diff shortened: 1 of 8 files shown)


PatilHarshh:
@Parvinmh please review

PatilHarshh:
#2104

src/packages/tour/components/TourTooltip.ts
```diff
@@ -217,6 +218,25 @@ const NextButton = ({
   return nextButton;
 };
 
+const ExitButton = ({
```
binrysearch, line 221: The cross (X) button in the header already provides the functionality to exit the tour. Could you please help me understand why we need a separate button for this?

From the UX point of view, adding more buttons to the footer of the dialog is not needed given that the X button already provides the same functionality.

I'm still happy to provide a way for users to customise the dialog though. That way, you can add/remove buttons to the dialog in your application when constructing the Intro.js Tour. This might require a bit of work on our side but then allows users to have fully customisable dialogs.

binrysearch (changes requested):

Issue #2107, opened 2025-06-27
Improve Accessibility to Meet WCAG Standards – Checklist


Faryad8:
To ensure Intro.js provides an accessible experience aligned with **WCAG guidelines**, we need to implement and verify the following checklists.

This issue collects both **UI/UX-focused** and **Development-focused** accessibility requirements to ensure all interactive elements in Intro.js tours are usable by all users, including those relying on screen readers or keyboard navigation.

**UI/UX Accessibility Checklist**
 
- [ ] Color contrast between text and background is sufficient in both normal and dark modes.
- [ ]  Buttons (Next, Skip, Done, etc.) are large enough and clearly clickable.
- [ ]  Instructions are clear, simple, and free of ambiguous terminology.
- [ ]  Only one central message per step is shown; users should not get distracted by multiple messages.
- [ ]  Buttons and tour elements display a visible focus style (e.g., outline) when navigated by keyboard (Tab).
- [ ]  Order of elements in the DOM and their visual presentation is predictable and consistent.
- [ ]  Close icon (×) is clearly visible and recognizable.
- [ ]  The tour remains readable and usable on mobile devices without requiring horizontal scrolling.
- [ ]  Animations are reduced or disabled when the user has the “reduce motion” preference enabled.

**Development Accessibility Checklist**
 
- [x] All buttons use `<button>`, or if `<a>` is used, they include `role="button"` and `tabindex="0"`.
- [x]  All interactive elements (buttons, bullets) have accessible names via `aria-label` or readable text.
- [x]  All interactive elements can be focused using the Tab key.
- [x]  The Tab order is logical and matches the visual flow of the tour.
- [x]  Focus automatically moves to the tooltip or first button when a step opens.
- [x]  The entire tour can be completed using keyboard-only navigation.
- [x]  `aria-live` or `aria-describedby` is used to notify screen readers of new content during the tour.
- [x]  When the tour ends, focus returns to the element that was focused before the tour started.
- [x]  `aria-hidden="true"` is applied to elements not relevant to the current step.

**Why this matters:**
Providing a fully accessible experience ensures that Intro.js can be used by everyone, including those who rely on assistive technologies. Meeting these accessibility requirements will make Intro.js compliant with WCAG 2.1 standards, improve usability, and broaden the audience who can benefit from interactive product tours.

Issue #1646, opened 2022-04-29
No accessible name for bullet (data-step) lists


digilou:
## Description
The unordered list of links to navigate from one tour step to the next doesn't create an accessible name for each link.

## Expected Behavior
An interactive component (i.e. hyperlinks, buttons) should have a role and name, according to [WCAG 4.1.2](https://www.w3.org/WAI/WCAG21/quickref/#name-role-value).
Accessibility expectations:
* A screen reader user should be presented with the purpose of this "button". Example: "Step 1" or {{ title of step }}
* A voice control user should be able to say the name of this component to activate it. (if they knew what the name was, although no visible label is available to them - )

## Actual Behavior
The links have been given the role of button. But no name (description of what this component does) isn't present (nor accessible). Testing with VoiceOver on macOS alerted me that the accessible name is missing.

## Errors and Screenshots (optional)
Visual of tour steps from a project under development:
<img width="293" alt="2 steps (gray dots) are visible and centered above the Back and Next buttons" src="https://user-images.githubusercontent.com/8865558/165872614-beb7301b-7225-4d5f-aadc-eaed032ad2b5.png">

Example rendered code in a project under development:
```
<div class="introjs-bullets">
  <ul role="tablist">
    <li role="presentation">
      <a role="button" class="" tabindex="0" data-step-number="1">&nbsp;</a>
    </li>
    <li role="presentation">
      <a role="button" tabindex="0" data-step-number="2" class="active">&nbsp;</a>
    </li>
  </ul>
</div>
```

## Example (recommended)
The project I'm working on is currently under development.

However, this can be tested even on [intro.js docs example page](https://introjs.com/docs/examples/basic/hello-world)

## Environment (optional)
| Tech    | Spec           |
----------|-----------------
| Browser | Firefox 99.0.1 |
| OS      | macOS 12.3.1   |
| Version | 5.1.0          |
| Frameworks | Rails, jQuery |

Issue #1116, opened 2020-12-13
Hints hotspot not accessible


namithubot:
## Description
Describe one (and only one) issue or enhancement you want to see in Intro.js (before writing any code).

## Expected Behavior
The hotspots (the one that triggers showing of hints upon clicking) should be accessible via tab button.

## Actual Behavior
Tab button is not focusing on the hotspots/triggers in spite of having `role="button" tabindex="0"`

## Example (recommended)
- Go to https://introjs.com/example/programmatic/hint.html
- Add hints.
- Try to access the hotspots/triggers via tab button.

## Environment (optional)
Firefox (78.5), Chrome (87.0.4280.88) -- Both 64 bit.

wouter-leistra:
We would love to get this fixed. We have some users with screen readers and we do want the pulsating hint markers to be accessible, navigational and interactive.

PR #2132, opened 2025-08-21
test(tour): add axe-jest for accessibility testing

Parvinmh:


src/packages/hint/hintItem.ts
```diff
@@ -77,9 +77,12 @@ export const fetchHintItems = (hint: Hint) => {
         hintAnimation = hintAnimationAttr === "true";
       }
 
+      const text = element.getAttribute(dataHintAttribute);
```
binrysearch, line 80: rename to hint

Parvinmh:
#1752 
#1646 
#1116 
#2135 
#2107

src/packages/tour/components/TourTooltip.ts
```diff
@@ -344,7 +346,12 @@ export const Header = ({
   const titleEl = TooltipContent({
     text: title,
     tooltipRenderAsHtml: renderAsHtml,
-    container: h1({ className: tooltipTitleClassName }),
+    container: h1({
+      className: tooltipTitleClassName,
+      role: "heading",
+      "aria-level": 1,
+      "aria-label": title || text,
```
binrysearch, line 353: The `aria-label` for title should be come from the `title` field only and if it's empty, it shouldn't be rendered (if we're rendering an empty h1 when title is not set, it's a bug and needs to be fixed)

src/packages/tour/mock.ts
```diff
@@ -23,6 +23,7 @@ export const appendMockSteps = (targetElement: HTMLElement = document.body) => {
     "Mock element second to last"
   );
   mockElementThree.setAttribute(dataStepAttribute, "10");
+  mockElementThree.setAttribute("aria-label", "Mock Heading");
```
binrysearch, line 26: Do we need this `aria-label` for the tests?

PR #2143, opened 2025-10-01
Campaign

Parvinmh:


src/packages/campaign/manager.ts
```diff
@@ -0,0 +1,426 @@
...
+      intro: step.intro,
+      element: step.element,
+      position: step.position,
+      scrollTo: step.scrollTo,
+      disableInteraction: step.disableInteraction,
+      tooltipClass: step.tooltipClass,
+    }));
+
+    // Create and configure tour
+    const tour = new Tour();
```
binrysearch: `executeCampaign` is currently tied to `Tour` only but what if we want to launch a `Hint`, or a banner in the future? This function shouldn't really care about what needs to be executed, instead, there needs to be a factory that takes the type of experience that needs to be launched (e.g. tour, hint, etc) and then launches it.

src/packages/campaign/manager.ts
```diff
@@ -0,0 +1,426 @@
...
+    tour.setOptions({
+      ...campaign.tourOptions,
+      steps: tourSteps,
+    });
+
+    // Set up campaign-specific callbacks
+    this.setupCampaignCallbacks(tour, campaign);
+
+    // Store active tour
+    this.activeTours.set(campaignId, tour);
```
binrysearch: Similar to the previous comment, we shouldn't maintain `activeTours` specifically, but rather an active experience type -- Also, how this `activeTours` is being used? what's the purpose?

src/packages/campaign/triggers.ts
```diff
@@ -0,0 +1,705 @@
...
+        }
+        break;
+
+      case "scroll_depth":
+        if (isScrollDepthTrigger(trigger)) {
+          this.setupScrollDepthTrigger(campaignId, trigger, callback);
+        }
+        break;
+
+      case "element_visible":
```
binrysearch: I really like these triggers

src/packages/campaign/triggers.ts
```diff
@@ -0,0 +1,705 @@
...
+      this.eventListeners.set(`${campaignId}-load`, () => {
+        DOMEvent.off(window, "load" as any, handler, false);
+      });
+    }
+  }
+
+  /**
+   * Setup scroll to element trigger
+   */
+  private setupScrollToElementTrigger(
```
binrysearch: These triggers should be moved to separate files to make the testing and maintainablitiy easier

src/packages/campaign/triggers.ts
```diff
@@ -0,0 +1,705 @@
...
+   */
+  private pauseIdleTimers(): void {
+    // Implementation for pausing idle timers
+  }
+
+  /**
+   * Resume idle timers when page becomes visible
+   */
+  private resumeIdleTimers(): void {
+    // Implementation for resuming idle timers
```
binrysearch: These need to be implemented I assume?

src/packages/campaign/types.ts
```diff
@@ -0,0 +1,345 @@
...
+  // Frequency and targeting
+  frequency?: CampaignFrequency;
+  targeting?: CampaignTargeting;
+
+  // Metadata
+  createdAt?: string;
+  updatedAt?: string;
+  author?: string;
+  tags?: string[];
+  priority?: number; // Priority for when multiple campaigns match (higher = higher priority)
```
binrysearch: Is this currently implemented?

src/packages/campaign/types.ts
```diff
@@ -0,0 +1,345 @@
...
+  triggers: CampaignTrigger[];
+
+  // Tour or Hint mode
+  mode: Tour | Hint;
+
+  // Tour configuration - uses existing TourOptions structure
+  tourOptions?: Partial<TourOptions>;
+
+  // Hint configuration - uses existing HintOptions structure
+  hintOptions?: Partial<HintOptions>;
```
binrysearch: I don't think we need to have tourOptions or hintOptions, but rather a generic "options" that defines the options of that particular experience based on what "mode" is passed in. Make sure we don't union the types though, there needs to be a BaseCampaign type and Campaign type the conditionally changes the "options" type based the "mode".

src/packages/campaign/types.ts
```diff
@@ -0,0 +1,345 @@
...
+/**
+ * Campaign collection (multiple campaigns)
+ */
+export interface CampaignCollection {
+  version: string;
+  campaigns: Campaign[];
+  global?: {
+    targeting?: CampaignTargeting;
+    tourDefaults?: Partial<TourOptions>;
+    hintDefaults?: Partial<HintOptions>;
```
binrysearch: Why do we need defaults here? And again, it shouldn't be tied to Tour or Hint.

src/packages/campaign/userTracker.ts
```diff
@@ -0,0 +1,104 @@
...
+  } | null = null;
+
+  /**
+   * Initialize the user tracker
+   */
+  async initialize(): Promise<void> {
+    if (this.isInitialized) return;
+
+    // Track session count
+    const sessionCountKey = "introjs-campaign-session-count";
```
binrysearch: This needs to be in configs.

example/campaigns/index.html
```diff
@@ -0,0 +1,237 @@
...
+      status.style.background = isError ? '#ffebee' : '#f0f0f0';
+    }
+
+    // Auto-load campaign on page load
+    window.addEventListener('DOMContentLoaded', async () => {
+      try {
+        updateStatus('Initializing campaigns from JSON file...');
+        
+        // Initialize campaigns from JSON file (correct path for this directory)
+        await introJs.initializeCampaigns("./campaigns/first-visit-tour.json");
```
binrysearch: How this would work in a browser environment? How do we do file I/O when running in browser context?

PR #2155, opened 2025-12-09
update cypress base snapshots

Parvinmh:


binrysearch:
Any idea why tour/dont-show-again.cy.ts/dont-show-again-clicked-first-step.png is not rendering the correct dialog?

PR #2195, opened 2026-08-07
fix(hint): remove resize listener in disableHintAutoRefresh

Parvinmh:


Parvinmh:
#2163 
@shaked-shlomo

Issue #2035, opened 2024-02-26
RTL keyboard arrows wrong


MordiSacks:
When in RTL the keyboard arrows are wrong, right should go back and left forward

PR #2198, opened 2026-09-08
fix(tour): reverse arrow key navigation in RTL tours

Parvinmh:
ArrowLeft/ArrowRight were hardcoded to previousStep/nextStep regardless of text direction. In RTL, this is backwards: right arrow should go back and left arrow should go forward.

Detect RTL by checking the computed direction of the rendered tooltip (matches introjs-rtl.css, which sets direction: rtl only on .introjs-tooltip), falling back to the tour's target element for pages that set dir="rtl" directly. Swap the arrow key behavior accordingly.


Parvinmh:
#2035

PR #2199, opened 2026-09-08
fix(tour): complete the tour on nextStep() when already on the last step

### LINKED ISSUE #951: TypeError when calling _nextStep function on last introItem by pressing right arrow key.

Parvinmh:
Fixes #951.

`Tour.isEnd()` checked `currentStep >= steps.length`, but `setCurrentStep()` already refuses to move past the last valid index, so that condition could never become true. As a result, calling `nextStep()` while already on the last step (e.g. via the right arrow key, or the public `nextStep()` API) silently did nothing instead of completing/exiting the tour the way clicking the Next/Done button does.

Fix `isEnd()` to be true once the current step is the last valid step (`>= steps.length - 1`), matching `isLastStep()`'s bounds and the button click handlers in TourRoot. This also fixes the same dead-code check in onKeyDown's Enter-on-skip-button handler, which had the identical bug.


Parvinmh:
#951

